PR #14, opened 2020-03-17
Rename to sysl-go and fix issues with make test

cuminandpaprika:
- Add gitignore
- Rename references to `sysl-go-comms` to `sysl-go`
- Fix missing parameter in `makefile`
- Fix tests so that `make gen` and `make test` run correctly

Note: This PR will be **squash merged**

PR #15, opened 2020-03-17
rename server-lib/sysl-go-comms to sysl-go

ChloePlanet:
`sysl` to brand it as part of the sysl toolkit,
`go` because it's specifically for Go codegen, not Java, Kotlin, etc.
It’s more an integrations library, we combine two things here: the runtime libraries and the code generator.

We removed `server` because it's not ultimately just for server codegen
(could be for a background job running under cron, or a cli client for downstreams) and
`lib` because it's only partly a library and it's fairly uncommon to append lib to libraries anyway.


ChloePlanet:
Dup #14.

PR #16, opened 2020-03-17
Configure http proxy for downstream.

springwiz:
Allow config to override the proxy config in the downstream http client.


gkanz (approved):
LGTM.

Issue #6, opened 2020-03-15
Setup CI using Github Actions


cuminandpaprika:
Setup Continuous Integration on this repo.

Requirements
- [x] Linting
    - Golangci-lint can be run several ways, either via reviewdog or codeclimate to get comments on the PR. Or it can be run directly as a check.
- [ ] Code Analysis
    - SonarCloud https://sonarcloud.io/dashboard?id=cuminandpaprika_sysl-go-comms
- [x] Run tests
- [x] Check Test coverage
- [x] Check go mod tidy doesn't modify any files
    - Something like https://blog.urth.org/2019/08/13/testing-go-mod-tidiness-in-ci/

Stretch Goals
- Spell check commit messages https://github.com/marketplace/actions/check-spelling
- Run and deploy example repo for preview, e.g https://zeit.co/github#features or heroku
- Automated Release Tagging
Blocked till https://github.com/anz-bank/sysl/issues/666 is closed
- Validate transform files with `sysl validate`
- Validate makefile keeps git state in porcelain state

cuminandpaprika:
Investigations into Coveralls seem to prove that Go support is secondary, and documentation seems to be outdated. Refer to https://docs.coveralls.io/go
- Go support is user contributed and the docs are for an older version of Github Actions

Running with the following config
```
    - name: Publish test results to Coveralls
      uses: coverallsapp/github-action@v1.0.1
      with: 
        github-token: ${{ secrets.github_token }}
        path-to-lcov: ./coverage.out
```

Produced these results:
```
[error] "2020-03-17T07:21:37.830Z"  'error from lcovParse: ' 'Failed to parse string'
[error] "2020-03-17T07:21:37.830Z"  'input: ' 'mode: set\n' +
```
https://github.com/cuminandpaprika/sysl-go-comms/pull/1/checks?check_run_id=513083763

cuminandpaprika:
Comparisons of SonarCloud Scanning methods:
- SonarCloud Automated Scanning [docs](https://sonarcloud.io/documentation/analysis/automatic-analysis/) 
    - Doesn't do code coverage 
- [Github Action](https://github.com/SonarSource/sonarcloud-github-action) 
    - Github Action is slightly flakey when posting results. [example](https://github.com/cuminandpaprika/sysl-go-comms/runs/515119261) on a push that should've had no impact on SonarCloud's config.
    - Enables Code Coverage
[...]

PR #23, opened 2020-03-18
Add Continuous Integration using Github Actions

cuminandpaprika:
Addresses https://github.com/anz-bank/sysl-go/issues/6

This PR implements some basic CI on the sysl-go repo
- Linting
   - It uses golangci-lint 1.23.8 for linting via the [Reviewdog Github Action](https://github.com/reviewdog/action-golangci-lint) as golangci.com is shutting down on the 15th April [blog post](https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e)
- Test Coverage
    - Runs `make check-coverage` to ensure test coverage is sufficient

Other steps explored in https://github.com/anz-bank/sysl-go/issues/6 are not addressed by this PR and will be added in subsequent PRs

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,37 @@
+name: Lint and Test
+on:
+  push:
+    branches:
+      - "*"
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Run golangci-lint 1.23.8
+        uses: actions-contrib/golangci-lint@v1
+        with:
+          golangci_lint_version: 1.23.8
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Run tests
+        run: make check-coverage
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,42 @@
+linters-settings:
+  lll:
+    line-length: 180
+
+linters:
+  enable-all: true
+  disable:
+    - dupl
+    - gochecknoglobals
+    - wsl
+    - lll
+    - gomnd # Ignore magic as it seems to produce too much noise for this codebase
+
+issues:
+  exclude-rules:
+    #Ignore variable scope linting
+    - path: "errors/error_handler_test.go"
+      linters:
+        - scopelint
+    - path: "config/config_test.go"
+      linters:
+        - scopelint
+    # Ignore things like "FormId should be FormID"
+    - path: codegen
+      text: "ST1003:"
+      linters:
+        - stylecheck
+    # Ignore things like "FormId should be FormID"
+    - path: codegen
+      text: ".*should be.*"
+      linters:
+        - golint
+    
+    - path: _test\.go
+      linters:
+        - govet # Ignore err shadowing in test files
+        - gomnd 
+        - funlen # Ignore function length for tests
+
+run:
+  deadline: 2m
+
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,52 @@
-COVERFILE=coverage.out
+all: test check-coverage lint tidy ## Tests, lints and checks coverage
 
+clean:  ## Remove generated files
+
+.PHONY: all clean
+
+# -- Lint ----------------------------------------------------------------------
+lint: ## Lint Go Source Code
+	golangci-lint run
+
+tidy: ## Run go mod tidy
+	go mod tidy
+
+.PHONY: lint tidy
+
+# -- Test ----------------------------------------------------------------------
+COVERFILE=coverage.out
+COVERAGE = 50
+
+test: ## Run all tests
+	go test -coverprofile=$(COVERFILE) -tags codeanalysis ./...
+
+check-coverage: test  ## Check that test coverage meets the required level
+	@go tool cover -func=$(COVERFILE) | $(CHECK_COVERAGE) || $(FAIL_COVERAGE)
+
+coverage: test  ## Show test coverage in your browser
+	go tool cover -html=$(COVERFILE)
+
+clean: 
+	rm -f $(COVERFILE)
+
+CHECK_COVERAGE = awk -F '[ \t%]+' '/^total:/ && $$3 < $(COVERAGE) {exit 1}'
+FAIL_COVERAGE = { echo '$(COLOUR_RED)FAIL - Coverage below $(COVERAGE)%$(COLOUR_NORMAL)'; exit 1; }
+
+.PHONY: check-coverage coverage test
+
+# --- Utilities ---------------------------------------------------------------
+COLOUR_NORMAL = $(shell tput sgr0 2>/dev/null)
+COLOUR_RED    = $(shell tput setaf 1 2>/dev/null)
+COLOUR_GREEN  = $(shell tput setaf 2 2>/dev/null)
+COLOUR_WHITE  = $(shell tput setaf 7 2>/dev/null)
+BOLD          = $(shell tput bold 2>/dev/null)
+
+help:
+	@awk -F ':.*## ' 'NF == 2 && $$1 ~ /^[A-Za-z0-9_-]+$$/ { printf "$(BOLD)$(COLOUR_WHITE)%-20s$(COLOUR_NORMAL)%s\n", $$1, $$2}' $(MAKEFILE_LIST) | sort
+
+.PHONY: help
+
+# -- Codegen ----------------------------------------------------------------------
 # Transform settings - common across all code generation
 TRANSFORMS=codegen/transforms
 GRAMMAR=codegen/grammars/go.gen.g
@@ -15,28 +62,8 @@ run-protoc=protoc --proto_path=$(PROTO_IN) --go_out=plugins=grpc:$(PROTO_OUT) $^
 
 GRPC_SERVER_FILES=grpc_interface.go grpc_handler.go
 
-tidy:
-	go mod tidy
+gen: ## Run sysl codegen and proto codegen
 
-gen:
-
-test: gen
-	go test -cover -tags codeanalysis ./...
-
-test-coverage:
-	go test -coverprofile=$(COVERFILE) ./...
-
-lint: gen
-	golangci-lint run
-
-cover: test-coverage
-	go tool cover -html=$(COVERFILE)
-
-cleanall: clean tidy
-
-clean:
-	go clean ./...
-
-.PHONY: gen test test-coverage lint cover clean cleanall tidy
+.PHONY: gen
 
 include codegen/testdata/*/Module.mk
--- a/README.md
+++ b/README.md
@@ -14,19 +14,30 @@ Go get the repository
 
 Ensure your environment provides:
 
-- [go 1.12.9](https://golang.org/)
-- [golangci-lint 1.17.1](https://github.com/golangci/golangci-lint)
-- some working method of obtaining dependencies listed in `go.mod` (working internet access, `GOPROXY`)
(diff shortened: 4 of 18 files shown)


.github/workflows/go.yml
```diff
@@ -0,0 +1,49 @@
...
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Get project dependencies
+        run: go get -v -t -d ./...
```
juliaogris: Same: `go get` is a pre go modules directive, you shouldn't have to run this.

.github/workflows/go.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Run tests
+        run: go test -cover -tags codeanalysis -coverprofile=coverage.out ./...
+
+      - name: Publish test results to Codecov
```
juliaogris: I think there is no need to publish this to Codecov. Do it with the makefile.

codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
+  precision: 2
+  round: down
+  range: "50...100"
+ignore:
+  - codegen/tests/*
```
juliaogris: EOL on EOF.

juliaogris: Here is why: https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b

juliaogris (commented):
Ok, I've looked at this.
I'd say, avoid unnecessary external systems, especially codecov/coverall and linting (not sure about sonarcube).
I'd also say use the Makefile in the github actions as well so you can simulate a CI run locally with make.

.github/workflows/go.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
+        id: go
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
```
juliaogris, line 34: Is it really necessary to checkout the code on every action "job".

cuminandpaprika:
> Ok, I've looked at this.
> I'd say, avoid unnecessary external systems, especially codecov/coverall and linting (not sure about sonarcube).
That's a good point, I've already run into some issues with some of the external services. I'll look into printing the codecoverage locally, thanks!

> I'd also say use the Makefile in the github actions as well so you can simulate a CI run locally with make.
Sounds good, definitely agree on the use of the Makefile to enable you to run as much as possible locally.

.golangci.yml
```diff
@@ -0,0 +1,69 @@
...
+  deadline: 2m
+  build-tags:
+    - codeanalysis
+
+# golangci.com configuration
+# https://github.com/golangci/golangci/wiki/Configuration
+service:
+  golangci-lint-version: 1.19.x # use the fixed version to not introduce new linters unexpectedly
+  prepare:
+    - echo "here I can run custom commands, but no preparation needed for this repo"
```
juliaogris: Remove the line above.

.golangci.yml
```diff
@@ -0,0 +1,69 @@
...
+        - scopelint
+    - path: "config/config_test.go"
+      linters:
+        - scopelint
+    # Ignore things like "FormId should be FormID"
+    - path: codegen
+      text: "ST1003:"
+      linters:
+        - stylecheck
+    # We like underscores
```
juliaogris: Really?

cuminandpaprika: haha I just copied this over from the existing .golangci.yml file in `server-lib`, so can't comment on the accuracy of liking underscores :)

codecov.yml
cuminandpaprika: Thanks, removed the codecov file

.github/workflows/go.yml
cuminandpaprika: I believe so, jobs run in parallel in separate environments so each one will need to checkout to code independently

Edit: Clearer explanation here https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#using-the-checkout-action

Docs here
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs

.golangci.yml
cuminandpaprika: Removed the entire service section, thanks!

.github/workflows/go.yml
cuminandpaprika: Yup, agreed, I've removed the dependency on Codecov (I think it just adds noise to the PR)

codecov.yml
cuminandpaprika: Fixed, thanks!

cuminandpaprika:
@juliaogris 

Here's what I've gone with instead:
- Update the makefile so that CI can run make commands
- Uses [this](https://github.com/actions-contrib/golangci-lint ) action to run golangci-lint
    - I didn't see a need to make our own, and this one allows the version of golangci-lint to be specified
    - Potentially, something like https://github.com/marketplace/actions/golangci-lint-action would be great, as it posts annotations back to github, which link to the line of code that needs to be fixed. But it doesn't support specifying a version of golangci-lint
- Remove codecov and run `make test-coverage` as suggested

- [x] Still need to fix the linting issues

.github/workflows/go.yml
juliaogris: Thanks.

.golangci.yml
```diff
@@ -0,0 +1,67 @@
...
+
+linters:
+  enable-all: true
+  disable:
+    - dupl
+    - gochecknoglobals
+    - gocyclo
+    - goconst
+    - lll
+    - misspell
```
juliaogris: that's a good one to have imo.

In fact I'd keep:
lll, misspell, gochecknoglobals, and gocyclo.
You can update gocyclo to higher values in gocyclo specific settings;
you can also "exclude" individual lines / files with //nolint:lll for instance

Makefile
```diff
@@ -1,5 +1,52 @@
...
+.PHONY: all clean
+
+# -- Lint ----------------------------------------------------------------------
+lint: ## Lint Go Source Code
+	golangci-lint run
+
+tidy:
+	go mod tidy
+
+.PHONY: lint tidy clean
```
juliaogris: no clean target here?

validator/validator_test.go
```diff
@@ -135,7 +135,7 @@ func testDummyValidator(reflect.Value) interface{} {
 	return 6
 }
 
-//nolint:gochecknoinits
+//nolint:gochecknoinits // We ignore these for tests
 func init() {
```
juliaogris: Maybe this should / could be done in [TestMain](https://golang.org/pkg/testing/#hdr-Main)?

README.md
```diff
@@ -14,19 +14,30 @@ Go get the repository
 
 Ensure your environment provides:
 
-- [go 1.12.9](https://golang.org/)
-- [golangci-lint 1.17.1](https://github.com/golangci/golangci-lint)
-- some working method of obtaining dependencies listed in `go.mod` (working internet access, `GOPROXY`)
-- env var `GOFLAGS="-mod=vendor"`
-- correctly configured cntlm or alpaca proxy running on localhost at port 3128 (only for updating vendor dependencies)
+- [go 1.13](https://golang.org/doc/install)
```
juliaogris, line 17: maybe time to upgrade to 1.14.1 ?

Makefile
```diff
@@ -1,5 +1,52 @@
...
+coverage: test  ## Show test coverage in your browser
+	go tool cover -html=$(COVERFILE)
+
+clean: 
+	rm -f $(COVERFILE)
+
+CHECK_COVERAGE = awk -F '[ \t%]+' '/^total:/ && $$3 < $(COVERAGE) {exit 1}'
+FAIL_COVERAGE = { echo '$(COLOUR_RED)FAIL - Coverage below $(COVERAGE)%$(COLOUR_NORMAL)'; exit 1; }
 
+.PHONY: check-coverage coverage test
```
juliaogris, line 35: `test` target missing.

Makefile
```diff
@@ -1,5 +1,52 @@
...
+
+# --- Utilities ---------------------------------------------------------------
+COLOUR_NORMAL = $(shell tput sgr0 2>/dev/null)
+COLOUR_RED    = $(shell tput setaf 1 2>/dev/null)
+COLOUR_GREEN  = $(shell tput setaf 2 2>/dev/null)
+COLOUR_WHITE  = $(shell tput setaf 7 2>/dev/null)
+BOLD          = $(shell tput bold 2>/dev/null)
+
+help:
+	@awk -F ':.*## ' 'NF == 2 && $$1 ~ /^[A-Za-z0-9_-]+$$/ { printf "$(BOLD)$(COLOUR_WHITE)%-20s$(COLOUR_NORMAL)%s\n", $$1, $$2}' $(MAKEFILE_LIST) | sort
```
juliaogris, line 45: I think it's great that you are re-using this `help` target.

Maybe also add help messages for all targets with `##`.
You can look at the output with `make help`

juliaogris (commented):
I think this is quite alright as is.
However, if you do agree with some comments and you don't want to update them as part of this PR, please write an issue for later.

It would also be great to have the git history cleaned up a bit.

.golangci.yml
cuminandpaprika: Removed this in the latest commits

Makefile
cuminandpaprika: Thanks for that, added help messages for all targets, it's a really nice feature to have in a makefile!

README.md
cuminandpaprika: I think I'd like to keep the version of this at least consistent with Sysl, given that it's the same team of developers. Will create an issue for it though https://github.com/anz-bank/sysl-go/issues/34

Makefile
cuminandpaprika: Thanks, fixed in 728d9a010242a1d6784cb6665c4f590129a3b23a

.golangci.yml
cuminandpaprika: That's a great point, 
I've updated to turn on `misspell`, `gocyclo` and `goconst`

It looks like it'd be quite a bit more work to enable `lll` and `gochecknoglobals` so I'll raise those as issues 
https://github.com/anz-bank/sysl-go/issues/33
https://github.com/anz-bank/sysl-go/issues/32

Makefile
cuminandpaprika: Sorry, a little bit confused by this. The target for `test` is above I think
```
test: ## Run all tests
	go test -coverprofile=$(COVERFILE) -tags codeanalysis ./...
```

validator/validator_test.go
cuminandpaprika: Good point, fixed in e97548cb87552a5b17071bd5097d429d82f62fe5

Makefile
juliaogris: yeah, so `.PHONY` needs to hold _all_ targets that don't match any actual file output generation. (this is an artefact from creating/ compiling c executables). Maybe we should talk about this at some point in more detail.

Anyway, long story short, you need to list `test` under `.PHONY` too.

juliaogris (commented):
This looks pretty good to me now, just needs that `.PHONY` update in the `Makefile`.
Also, would be great if you could squash your fixup commits and maybe just have three commits:
- linter updates
- makefile extension & docs update
- docs version updates
- CI
Or so....

Makefile
cuminandpaprika: Awesome, thanks for the explanation, `test` has been included undernder `.PHONY` in https://github.com/anz-bank/sysl-go/pull/23/commits/a459be5ee642cd9884c84430384c4e82f9e0abdd

cuminandpaprika:
> This looks pretty good to me now, just needs that `.PHONY` update in the `Makefile`.
> Also, would be great if you could squash your fixup commits and maybe just have three commits:
> 
> * linter updates
> * makefile extension & docs update
> * docs version updates
> * CI
>   Or so....

Ah, thanks for the example, I wasn't sure what granularity is desirable for commits, that helps a lot to guide me!

Rebased and squashed, and addressed your comment on the makefile target 👍

juliaogris (approved):

PR #31, opened 2020-03-19
Feat/handle error transform

### LINKED ISSUE #1: Move error handler to error mapper
Make `sysl-go-comms` responsible for handling (i.e. writing) errors, not just reporting them to custom code. 

In the current interaction pattern, `sysl-go-comms` requests that custom code handle errors. This requires that custom code map an error to an `HTTPError`, then write the `HTTPError` out:

```go
func HandleError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error) {
    // map error to http error
    // write http error
}
```

The replacement interaction pattern moves the responsibility of writing the `HTTPError` into `sysl-go-comms` and keeping custom code responsible for mapping the error: 

```go
func MapError(ctx context.Context, kind common.Kind, message string, cause error) common.HTTPError {
    // return mapped http error
}
```

tiagralee:
issue #1 
- Add a common.CustomError
- Changed `svc_handler.sysl` to transform the error type to the CustomerError map
- Changed `svc_handler.sysl` add a new function HandleError to map custom error, if not found return the common http error.
- Update the tests for simple and dbendpoints.


gkanz:
@tiagralee, could you please add a summary to the PR regarding why you chose to approach you did to meet the requirement given in #1? There are no commit comments for me to go on...

tiagralee:
@gkanz Thanks Greg, just add a summary.

codegen/tests/dbendpoints/dbendpoints_gencallback.go
```diff
@@ -30,12 +30,10 @@ func (c Callback) Config() validator.Validator {
 	return Config{}
 }
 
-func (c Callback) HandleError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error) {
+func (c Callback) MapError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error) common.HTTPError {
```
andrewemeryanz: `MapError` shouldn't get access to a `ResponseWriter`.

codegen/tests/dbendpoints/servicehandler.go
```diff
@@ -34,10 +35,16 @@ func NewServiceHandler(genCallback GenCallback, serviceInterface *ServiceInterfa
 	return &ServiceHandler{genCallback, serviceInterface, db}
 }
 
+// Handler Error
+func (s *ServiceHandler) HandleError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error) {
```
andrewemeryanz: `HandleError` can be private: `handleError`

andrewemeryanz (changes requested):
Minor changes requested.

gkanz:
I have some questions about the approach: I've added a comment to https://github.com/anz-bank/sysl-go/issues/1 with my thoughts.

tiagralee:
Thanks Greg, I'll have a look.

codegen/tests/dbendpoints/servicehandler.go
tiagralee: Thanks Andrew, just make this lowercase will make it private?

andrewemeryanz: Yes.

codegen/tests/simple/servicehandler.go
```diff
@@ -35,10 +42,32 @@ func NewServiceHandler(genCallback GenCallback, serviceInterface *ServiceInterfa
 	return &ServiceHandler{genCallback, serviceInterface}
 }
```
tiagralee, line 46: As the current transformation does not support switch statement, use ifelse statement instead, the below code equivalent to `
serverError := common.CreateError(ctx, kind, message, cause)
	httpError := s.genCallback.MapError(ctx, serverError)
	if httpError == nil {
		switch t := cause.(type) {
                 case CustomError:
                      e := t.HTTPError()
                      httpError = &e
                 default:
                      e := common.HandleError(ctx, serverError)
                      httpError = &e
                }
	}
	httpError.WriteError(ctx, w)
`

codegen/tests/simple/servicehandler.go
```diff
@@ -11,6 +12,12 @@ import (
 	"github.com/anz-bank/sysl-go/validator"
 )
 
+// *BusinessLogicError error
+var businesslogicerror = map[string]string{"name": "BusinessLogicError", "http_code": "1001", "http_message": "foo", "http_status": "500"}
+
+// *BusinessLogicError2 error
+var businesslogicerror2 = map[string]string{"name": "BusinessLogicError2", "http_code": "1002", "http_message": "foo2", "http_status": "501"}
```
andrewemeryanz: `businesslogicerror2` needs to be capitalised to be public: 
```go
var BusinessLogicError2 = ...
```

Please make this value be the name of the type with a capital letter at the start:
```text
!type FooBar [~error]: ...     # becomes: val FooBar = ...
!type foobar [~error]: ...     # becomes: val Foobar = ...
```

codegen/tests/simple/simple_gencallback.go
```diff
@@ -27,12 +26,9 @@ func (c Callback) Config() validator.Validator {
...
 }
 
-func (c Callback) HandleError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error) {
-	se := common.CreateError(ctx, kind, message, cause)
-
-	httpError := common.HandleError(ctx, se)
-
-	httpError.WriteError(ctx, w)
+func (c Callback) MapError(ctx context.Context, cause error) *common.HTTPError {
+	handleError := common.HandleError(ctx, cause)
```
andrewemeryanz: You can just return `nil` here. `common.HandlerError` is called by the `handlerError` function internally.

codegen/tests/simple/types.go
```diff
@@ -16,6 +16,13 @@ var _ = time.Parse
 // Reference imports to suppress unused errors
 var _ = date.Parse
 
+// BusinessLogicError ...
+type BusinessLogicError struct {
```
andrewemeryanz: We shouldn't be generating types here for items tagged with `[~error]`.

common/custom_error.go
```diff
@@ -0,0 +1,30 @@
...
+type CustomError map[string]string
+
+func (e CustomError) Error() string {
+	return fmt.Sprintf("%s(%v)", e["name"], e)
+}
+
+func (e CustomError) HTTPError() HTTPError {
+	httpStatus, err := strconv.Atoi(getOrDefault(e, "http_status", "500"))
+	if err != nil {
+		httpStatus = 500
```
andrewemeryanz, line 22: We should be logging this error. Pass in the `Context` if necessary.

codegen/tests/simple/servicehandler.go
```diff
@@ -35,10 +42,32 @@ func NewServiceHandler(genCallback GenCallback, serviceInterface *ServiceInterfa
 	return &ServiceHandler{genCallback, serviceInterface}
 }
 
+// Handler Error
+func (s *ServiceHandler) handleError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error) {
```
andrewemeryanz: This function doesn't need to be generated because it doesn't contain any custom code. The following code could replace `common.HandleError`:

```go
func HandleError(ctx context.Context, w http.ResponseWriter, kind Kind, message string, cause error,
		httpErrorMapper func(context.Context, ServerError) *HTTPError) {

	logEntry := GetLogEntryFromContext(ctx)
	logEntry.Error(cause)

	err := CreateError(ctx, kind, message, cause)
	httpError := httpErrorMapper(ctx, err)
	if httpError == nil {
		switch t := cause.(type) {
		case CustomError:
			e := t.HTTPError()
			httpError = &e
		default:
			e := mapError(ctx, &err)
			httpError = &e
		}
	}
	httpError.WriteError(ctx, w)
}

func mapError(ctx context.Context, err error) HTTPError {

	var (
		httpCode        int
		errorCode, desc string
	)
```

andrewemeryanz (changes requested):
We're getting closer. We can avoid a bunch of generated code by doing more in `sysl-go`.

common/custom_error.go
tiagralee: The only reference is from the HandleError, which log the error before calling this.

common/error_handler.go
```diff
@@ -13,7 +14,27 @@ const (
...
+			httpError = &e
+		default:
+			e := MapError(ctx, err)
+			httpError = &e
+		}
+	}
+	httpError.WriteError(ctx, w)
+}
+
+func MapError(ctx context.Context, err error) HTTPError {
```
tiagralee, line 37: Need to make the MapError public, as the defaultcallback need to use this.

common/custom_error.go
andrewemeryanz: Sorry, I mean the `err` returned as a result of `strconv.Atoi`.

common/error_handler.go
andrewemeryanz: > Need to make the MapError public

Yes. That's fine.

common/custom_error.go
andrewemeryanz: I was thinking something more like:
```
func (e CustomError) HTTPError(ctx context.Context) HTTPError {
	httpStatusString := getOrDefault(e, "http_status", "500")
	httpStatus, err := strconv.Atoi(httpStatusString)
	if err != nil {
		logEntry := GetLogEntryFromContext(ctx)
		logEntry.Error(fmt.Sprintf("invalid http_status: %s for error: %s, returning 500", httpStatusString, e["name"]))
		httpStatus = 500
	}
```

Now `500` should really be a constant somewhere in `sysl-go` but let's park that for now. Feel free to put a `TODO` in there for now.

andrewemeryanz (changes requested):
One minor change, then we'll tidy up the commit history to make it more readable.

PR #36, opened 2020-03-20
Added sysl time codegen tag differentiator

### LINKED ISSUE #35: Support different time marshalling formats in code generator
Currently the code generator only supports unmarshalling time strings to the inbuilt convert.JSONTime struct. Request for unmarshalling into the golang stdlib time.Time

nydrani:
Closes #35 

Adds a tag differentiator for sysl time data type. By default it will generate a DATETIME primitive type to convert.JSONType as the data format. Adding a tag: @time_format = "stdtime" will code generate to time.Time

Not sure where to put the docs for this??

### DIFF
--- a/codegen/transforms/svc_types.sysl
+++ b/codegen/transforms/svc_types.sysl
@@ -26,7 +26,7 @@ CodeGenTransform:
           "STRING_8" => "string"
           "BOOL" => "bool"
           "DATE" => "date.Date"
-          "DATETIME" => "convert.JSONTime"
+          "DATETIME" => if "time_format" in t.attrs then if "stdtime" == t.attrs.time_format then "time.Time" else "convert.JSONTime" else "convert.JSONTime"
         "sequence" => "[]" + GoType(t.sequence).out
         "set" => GoType(t.set).out + "Set"
         else GoName(t.type_ref).out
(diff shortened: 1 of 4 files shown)


codegen/transforms/svc_types.sysl
```diff
@@ -26,7 +26,7 @@ CodeGenTransform:
           "STRING_8" => "string"
           "BOOL" => "bool"
           "DATE" => "date.Date"
-          "DATETIME" => "convert.JSONTime"
+          "DATETIME" => if "time_format" in t.attrs then if "stdtime" == t.attrs.time_format then "time.Time" else "convert.JSONTime" else "convert.JSONTime"
```
springwiz, line 29: Use && to combine the first two conditions if "time_format" in t.attrs then if "stdtime" == t.attrs.time_format.

nydrani: I cant use && because the && operator doesnt short circuit in sysl. So using && will result in a runtime error.

springwiz: I think it should, please open an issue in sysl repo, if it doesn't.

nydrani:
https://github.com/anz-bank/sysl/issues/719

springwiz (approved):
LGTM.

PR #37, opened 2020-03-20
Hide config types from custom code

### LINKED ISSUE #3: Hide config types from custom code
Configuration for `sysl-go` is managed with the following process:
1. Create a `config.yaml` file that contains:
a. The `library` section for `sysl-go` library config
b. The `genCode` section for generated code configuration
c. Other sections read by custom code as required. 
2. Create a `Config` struct (in custom code) that contains yaml tags for the above sections.
3. Read the `config.yaml` file into the `Config` struct.
4. Use the `Config` internally.

The goal of this piece of work is to eliminate step 3 (above) by moving (into `sysl-go`) or generating (using `sysl-go` generators) as much of the configuration as possible.

The eventual goal of this work is to simplify the starting of a server to a command similar to the one below, where custom code knows only about its own configuration:

```go
gen.StartServer(
    func(ctx context.Context, config MyConfig) (foo.ServiceInterface, foo.GenCallback) {
        ...
    }
)
```

The above approach requires that `sysl-go` know how custom code provides the location of the configuration file. The config file should eventually be retrieved from the `config` argument on the command-line, however, to break this work into stages, the first step would be to provide a function that reads from a single config file and populates both custom and library config structs:

```go
func ReadConfig(file string, custom *interface{}) (common.Library, common.GenCode)
```
With usage:
```go
config := MyConfig{}
lib, gen := common.ReadConfig(file, &config)
```

ChloePlanet:
Fixed #3.

Changes proposed in this pull request:
- [x] func ReadConfig: reads from a single config file and populates both custom, library and genCode config structs
- [x] auto-generate downstream config struct and populate it


config/config.go
```diff
@@ -1,44 +1,49 @@
...
-type LibraryConfig struct {
-	Log       LogConfig `yaml:"log"`
-	Profiling bool      `yaml:"profiling"`
-}
+// ReadConfig reads from a single config file and populates both custom, library and genCode config structs
+// cfgFile: path to config file
+// config: a pointer to the custom config struct
+func ReadConfig(cfgFile string, lib *LibraryConfig, gen *GenCodeConfig, config interface{}) {
+	if cfgFile != "" {
+		viper.SetConfigFile(cfgFile)
```
andrewemeryanz: Can we do this without `viper` by parsing the yaml contents directly. `viper` is an unnecessary dependency if it can be done without it simply.

config/config_test.go
```diff
@@ -2,156 +2,59 @@ package config
 
 import (
 	"testing"
+	"time"
 
-	"github.com/anz-bank/sysl-go/common"
-	"github.com/anz-bank/sysl-go/validator"
-
+	"github.com/alecthomas/assert"
```
andrewemeryanz: Please replace with `github.com/alecthomas/assert` with `github.com/stretchr/testify/require` to keep consistent with other tests.

config/testdata/config.yaml
```diff
@@ -0,0 +1,57 @@
...
+      writeTimeout: 120s
+      common:
+        hostName: ""
+        port: 8080
+    grpc:
+      hostName: ""
+      port: 8081
+  downstream:
+    contextTimeout: 120s
+    fenergo:
```
andrewemeryanz: This really needs to be made anonymous, meaning `fenergo` should be renamed to `foo` or similar and urls need to changed.

config/testdata/config.yaml
```diff
@@ -0,0 +1,57 @@
...
+    http:
+      basePath: /admintest
+      readTimeout: 1s
+      writeTimeout: 3s
+      common:
+        hostName: ""
+        port: 3333
+genCode:
+  handlers:
+    getV1BffFenergoAPIDocs:
```
andrewemeryanz: Remove any unnecessary config: `getV1StatementOfPositionCustomer` isn't used outside this file.

andrewemeryanz (changes requested):
Approach and testing look good. Couple of necessary changes though.

anzdaddy (approved):

Makefile
```diff
@@ -21,7 +21,7 @@ COVERFILE=coverage.out
 COVERAGE = 50
 
 test: ## Run all tests
-	go test -coverprofile=$(COVERFILE) -tags codeanalysis ./...
+	go test -count=1 -cover -coverprofile=$(COVERFILE) -tags codeanalysis ./...
```
andrewemeryanz: This change could probably be a separate pull request.

andrewemeryanz (approved):
Looks good.

PR #39, opened 2020-03-23
Add restlib test

gkanz:
Tests for correct behaviour when parsing malformed JSON. Also cleaned up some test names and linting complaints.


springwiz (approved):
LGTM

PR #40, opened 2020-03-23
Support XML and PDF in SendHTTPResponse

jecofang01:
As needed for dcx/bff

### DIFF
--- a/restlib/restlib.go
+++ b/restlib/restlib.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"encoding/xml"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


springwiz:
@jecofang01 There was a similar to change I had done before to skip json encoding for xml content type in request and response while talking to downstream. All you need to do is to the set the content-type header in the context. Can you please let me know, how is your change different from that? Please refer to the test case func TestSendHTTPResponseXMLBody(t *testing.T) { in the rest lib. I can't add a reference to that PR as it was in the Server-lib repo.

jecofang01:
hi @springwiz , this change is for upstream. The generated code send out the response to upstream by calling `SendHTTPResponse`.

springwiz (dismissed):
Can you please show how the changes look like in the Sysl file. Please make the changes in the simple.sysl, generate go code and add the test cases to simple_test.go.

jecofang01:
@springwiz Thanks for your comments. I'm trying to add the test cases, however I was blocked by https://github.com/anz-bank/sysl/issues/725. I will update once the issue fixed.

ghost (approved):

PR #41, opened 2020-03-23
Add go mod tidy check

cuminandpaprika:
A common issue for go developers is they forget to run go mod tidy before committing a change. This results in unused references being left in the `go.mod` file.

This PR adds a check for go mod tidiness as suggested in https://github.com/anz-bank/sysl-go/issues/6#issuecomment-601015906


nydrani (approved):
LGTM

PR #42, opened 2020-03-24
Generate correct http clients in gRPC gen servers

gkanz:
The clients interfaces used by the http-generated code was updated a while ago however the gRPC server code was not updated. This change updates the clients generated for use within gRPC servers.

The main change is that a special request type is generated for every method.


nydrani (approved):
LGTM

PR #43, opened 2020-03-25
Use @go_package in preference to @package for svc_client.sysl svc_types.syls and svc_error_types.sysl

gkanz:
Tiny change to preferntially use @go_package in the package clause. This allows the generated code to be placed in a different location to that specified in @package.

Some of the other transforms are already doing this.


nydrani (approved):
LGTM

PR #46, opened 2020-03-27
Autogenerate app and wiring.

springwiz:
The intent of this PR is to autogenerate the app. This includes the autogeneration of:
1. Autogeneration of  LoadServices in app. This includes most of the wiring code.
2. Autogeneration of handlermanager in app.

Dependencies:
1. Autogenerated config.
2. Changes in the simple.sysl and deps.sysl.
3. Autogenerated ErrorHandler.


config/context.go
```diff
@@ -0,0 +1,33 @@
+package config
```
cuminandpaprika: This file doesn't seemed to be called anywhere, from what I can see? No description of it in the PR also, if it's needed, can you add a description of what this is used for in the commit message/ PR?

codegen/tests/simple/skeletonhandlerimpl.go
```diff
@@ -0,0 +1,58 @@
+// Code generated by sysl DO NOT EDIT.
```
cuminandpaprika: nitpick: I think perhaps naming the file `skeletonhandler.go` would be cleaner and more go-like?

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// LibraryConfig ...
+func (h *HandlerManager) LibraryConfig() *config.LibraryConfig {
+	return &h.coreCfg.Library
+}
+
+// LoadServices ...
+func (h *HandlerManager) LoadServices(ctx context.Context) error {
+	bffSI := ServiceInterface{
```
cuminandpaprika: bffSI should be replaced with a generic name
```suggestion
	appSI := ServiceInterface{
```

cuminandpaprika (commented):
Looks good so far, struggling to see the what the final user interface will look like though, so will take a look once the error types are merged in.

Just a note, if some of the generated methods in app.go will be called by the user, we might need some longer descriptions in the comments

codegen/tests/simple/skeletonhandlerimpl.go
```diff
@@ -0,0 +1,58 @@
...
+package simple
```
andrewemeryanz: What's the benefit of generating this file? Custom code doesn't need to provide a handler for every endpoint and every handler they do provide will be custom.

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,89 @@
...
+
+	genCallbacks := config.Callback{
+		UpstreamTimeout:   h.coreCfg.GenCode.Upstream.ContextTimeout,
+		DownstreamTimeout: h.coreCfg.GenCode.Downstream.(*DownstreamConfig).ContextTimeout,
+		ErrorHandler:      config.ErrorHandler{},
+		RouterBasePath:    "/",
+		UpstreamConfig:    h.coreCfg.GenCode.Upstream,
+	}
+
+	depsHTTPClient, depsErr := core.BuildDownstreamHTTPClient("deps", &h.coreCfg.GenCode.Downstream.(*DownstreamConfig).Deps)
```
andrewemeryanz: To keep changes small, the section below (to the end of the function) should be split out into a separate method and this should be the focus of the first pull request. Generated code should be something like:

```go
func InitialiseHandlers(handlerManager *HandlerManager, serviceInterface ServiceInterface, callbacks GenCallbacks) error {
    // build the client
    // build the service handler
    // build the service router
    // enable the handlers
}
```


The service interface will always be custom code, never generated. The construction of the callbacks will also be custom at this stage. So custom code can look something like:

```go
func (h *HandlerManager) LoadServices(ctx context.Context) error {
    serviceInterface := ...
    serviceCallback := ...
    error := InitialiseHandler(h, serviceInterface, serviceCallback)
    ...
}
```

andrewemeryanz (changes requested):
Let's reduce the scope of the first set of changes to only handler initialisation. Other parts can be generated at a later stage.

codegen/tests/simple/app.go
springwiz: Removing handler manager from the code and just generating the above portion would break the build, how do we plan to handle that?

codegen/tests/simple/skeletonhandlerimpl.go
springwiz: will be removed, its not needed anymore.

codegen/tests/simple/app.go
springwiz: Plan to remove it.

codegen/tests/simple/app.go
springwiz: Included the HandlerManager struct in the app, so that build will go through.

codegen/tests/simple/skeletonhandlerimpl.go
springwiz: removed.

config/context.go
springwiz: removed.

springwiz:
> Let's reduce the scope of the first set of changes to only handler initialisation. Other parts can be generated at a later stage.

Done. Please review.

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,34 @@
...
+// HandlerManager for Simple
+type HandlerManager struct {
+	coreCfg             *config.DefaultConfig
+	enabledHandlers     []handlerinitialiser.HandlerInitialiser
+	enabledGrpcHandlers []handlerinitialiser.GrpcHandlerInitialiser
+}
+
+// LoadServices ...
+func (a *App) InitialiseHandler(h *HandlerManager, serviceInterface ServiceInterface, serviceCallback core.RestGenCallback) error {
+	depsHTTPClient, depsErr := core.BuildDownstreamHTTPClient("deps", &h.coreCfg.GenCode.Downstream.(*DownstreamConfig).Deps)
```
andrewemeryanz: "deps" should be the name of the system (or some other meaningful identifier).

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,34 @@
...
+	if depsErr != nil {
+		return depsErr
+	}
+
+	depsClient := deps.NewClient(depsHTTPClient, h.coreCfg.GenCode.Downstream.(*DownstreamConfig).Deps.ServiceURL)
+	serviceHandler := NewServiceHandler(serviceCallback, &serviceInterface, depsClient)
```
andrewemeryanz: You'll need to handle multiple clients here also.

andrewemeryanz (dismissed):
Early feedback. I'll keep reviewing and get back to you if I have more.

codegen/tests/simple/app.go
springwiz: "deps" is the app name picked up from the model.

codegen/tests/simple/app.go
springwiz: The code is tested with bff already. But will see if I can add another downstream.

springwiz: Added a new downstream.

codegen/tests/simple/app.go
andrewemeryanz: The `App` isn't doing much here. Can we change this to a function instead:
```go
type HandlerInit struct {
    HttpHandlers []handlerinitialiser.HandlerInitialiser
    GrpcHandlers []handlerinitialiser.GrpcHandlerInitialiser
}

func InitialiseHandlers(coreCfg *config.DefaultConfig, serviceInterface *ServiceInterface, serviceCallback core.RestGenCallback) (*HandlerInit, error) {
    ...
    serviceRouter := ...
    httpHandlers := []handlerinitialiser.HandlerInitialiser{serviceRouter}
    return &HandlerInit{HttpHandlers: httpHandlers}, nil
}
```

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if downstreamErr != nil {
+		return downstreamErr
+	}
+
+	depsClient := deps.NewClient(depsHTTPClient, h.coreCfg.GenCode.Downstream.(*DownstreamConfig).Deps.ServiceURL)
+	downstreamClient := downstream.NewClient(downstreamHTTPClient, h.coreCfg.GenCode.Downstream.(*DownstreamConfig).Downstream.ServiceURL)
+	serviceHandler := NewServiceHandler(serviceCallback, &serviceInterface, depsClient, downstreamClient)
+	serviceRouter := NewServiceRouter(serviceCallback, serviceHandler)
+	h.enabledHandlers = append(h.enabledHandlers, serviceRouter)
```
andrewemeryanz: We also need to add tests to verify that the method is doing what it says it is for both http and grpc handlers, namely that the returned `HandlerInit` contains the expected http and grpc handlers that contain the expected router that contains the expected handlers, etc.

codegen/tests/simple/app.go
springwiz: The intent is have App is to have more content into it, in the future.

andrewemeryanz: Then let's make that change in the future when we know what it should look like.

codegen/tests/simple/app.go
springwiz: Will add.

codegen/tests/simple/app.go
springwiz: I can add this to handlerManger, that should be easy. Don't think there is much value in having another type.

andrewemeryanz: As a name, `HandlerManager` is ambiguous. This function doesn't need to operate on an object because the only purpose of the object would be to invoke its single method. 

Make it a function for now. `InitialiseHandlers` returns the initialised handlers.

springwiz: ok

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/anz-bank/sysl-go/codegen/tests/deps"
+	"github.com/anz-bank/sysl-go/codegen/tests/downstream"
+	"github.com/anz-bank/sysl-go/config"
+	"github.com/anz-bank/sysl-go/core"
+	"github.com/anz-bank/sysl-go/handlerinitialiser"
+)
+
+// HandlerInit for Simple
+type HandlerInit struct {
+	enabledHandlers     []handlerinitialiser.HandlerInitialiser
```
andrewemeryanz: Can these renamed to:
```go
RestHandlers []handlerinitialiser.HandlerInitialiser // capitalised for export
GrpcHandlers []handlerinitialiser.GrpcHandlerInitialiser
```

We'll be pushing for some naming consistency in the codebase soon.

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,37 @@
...
+	enabledGrpcHandlers []handlerinitialiser.GrpcHandlerInitialiser
+}
+
+// InitialiseHandler ...
+func InitialiseHandler(coreCfg *config.DefaultConfig, serviceInterface ServiceInterface, serviceCallback core.RestGenCallback) (*HandlerInit, error) {
+	var err error = nil
```
andrewemeryanz, line 20: This value isn't used except to return nil at the end of the function. Is this deliberate?

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// HandlerInit for Simple
+type HandlerInit struct {
+	enabledHandlers     []handlerinitialiser.HandlerInitialiser
+	enabledGrpcHandlers []handlerinitialiser.GrpcHandlerInitialiser
+}
+
+// InitialiseHandler ...
+func InitialiseHandler(coreCfg *config.DefaultConfig, serviceInterface ServiceInterface, serviceCallback core.RestGenCallback) (*HandlerInit, error) {
```
andrewemeryanz: `InitialiseHandlers`

codegen/tests/simple/app.go
```diff
@@ -0,0 +1,37 @@
...
+	if downstreamErr != nil {
+		return nil, downstreamErr
+	}
+
+	depsClient := deps.NewClient(depsHTTPClient, coreCfg.GenCode.Downstream.(*DownstreamConfig).Deps.ServiceURL)
+	downstreamClient := downstream.NewClient(downstreamHTTPClient, coreCfg.GenCode.Downstream.(*DownstreamConfig).Downstream.ServiceURL)
+	serviceHandler := NewServiceHandler(serviceCallback, &serviceInterface, depsClient, downstreamClient)
+	serviceRouter := NewServiceRouter(serviceCallback, serviceHandler)
+	httpHandlers := []handlerinitialiser.HandlerInitialiser{serviceRouter}
+	return &HandlerInit{enabledHandlers: httpHandlers}, err
```
andrewemeryanz: The types are fine for now. Some thought will need to be in to supporting both REST and GRPC soon.

codegen/tests/simple/app.go
springwiz: Yes, workaround to make codegen work.

codegen/tests/simple/app.go
springwiz: ok

codegen/tests/simple/app.go
springwiz: ok

anzdaddy (approved):

PR #50, opened 2020-03-27
make build metadata opinionated

### LINKED ISSUE #4: Move build metadata into sysl-go
Make build metadata opinionated and remove the need for custom code developers re-specify the same metadata variables in every server.

Do this by moving common build data into the `sysl-go-comms` library.

```go
var (
         // Build metadata. These values are intended to be overridden by values
         // supplied by the build environment at link time. Values are specified
         // by passing "-X package.Varname=Value" to the linker. For example:
         //
         //  go build -ldflags="-X main.Name=foo -X main.Version=1.0" ...etc...
         //
         // Ref: https://golang.org/cmd/link/
 
         // Name is set at the build time
         Name = ""
 
         // Version is set at the build time
         Version = ""
 
         // BuildID is set at the build time
         BuildID = ""
 
         // CommitSha is set at the build time
         CommitSha = ""
 
         // BranchName is set at the build time
         BranchName = ""
 
         // TagName is set at the build time
         TagName = ""
)
```

tiagralee:
issue #4 

- Add build metadata to the core and set default value in build time
- Removed the previous buildmetadata param in server initialise

### DIFF
--- a/core/build_metadata.go
+++ b/core/build_metadata.go
@@ -1,0 +1,40 @@
+package core
+
+import "github.com/anz-bank/sysl-go/status"
+
+var (
+	// Build metadata. These values are intended to be overridden by values
+	// supplied by the build environment at link time. Values are specified
+	// by passing "-X package.Varname=Value" to the linker. For example:
+	//
+	//  go build -ldflags="-X main.Name=TestApp -X main.Version=1.0" ...etc...
+	//
+	// Ref: https://golang.org/cmd/link/
+
+	// Name is set at the build time
+	Name = ""
+
+	// Version is set at the build time
+	Version = ""
+
+	// BuildID is set at the build time
+	BuildID = ""
+
+	// CommitSha is set at the build time
+	CommitSha = ""
+
+	// BranchName is set at the build time
+	BranchName = ""
+
+	// TagName is set at the build time
+	TagName = ""
+)
+
+var buildMetadata = &status.BuildMetadata{
+	Name:       Name,
+	Version:    Version,
+	BuildID:    BuildID,
+	CommitSha:  CommitSha,
+	BranchName: BranchName,
+	TagName:    TagName,
+}
--- a/core/server.go
+++ b/core/server.go
@@ -5,13 +5,12 @@ package core
 import (
 	"context"
 
-	"github.com/anz-bank/sysl-go/status"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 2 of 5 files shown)


core/server.go
```diff
@@ -5,13 +5,12 @@ package core
 import (
 	"context"
 
-	"github.com/anz-bank/sysl-go/status"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
 //nolint:gocognit // Long method names are okay because only generated code will call this, not humans.
-func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, logger *logrus.Logger, promRegistry *prometheus.Registry, buildMetadata *status.BuildMetadata) error {
```
andrewemeryanz, line 14: This parameter can also be removed from `core.configureAdminServerListener`.

andrewemeryanz (changes requested):
One minor change, otherwise fine.

core/build_metadata.go
```diff
@@ -0,0 +1,40 @@
+package core
+
+import "github.com/anz-bank/sysl-go/status"
+
+var (
+	// Build metadata. These values are intended to be overridden by values
+	// supplied by the build environment at link time. Values are specified
+	// by passing "-X package.Varname=Value" to the linker. For example:
+	//
+	//  go build -ldflags="-X main.Name=BFF -X main.Version=1.0" ...etc...
```
cuminandpaprika: Could we replace this example with something more generic?
e.g BFF isn't very clear as an application name. Something like `TestApp` or `DemoApp` could be more appropriate?

andrewemeryanz:
Looks good. Can you squash these changes into a single commit:
```sh
git reset --soft HEAD~3
git commit
```

and put a more descriptive commit message, something like:

```
Add build metadata

This change adds the ability to set build metadata at build time. Build
metadata is regarded as data that is common to all server builds (name, 
version, etc.).
```

cuminandpaprika (commented):
Looks good! 

One more thing that would be super helpful for any users is to document the need to run go build with the -ldflags, probably somewhere on the main sysl repo. If you don't want to jump on that right now, could you create an issue for it?

andrewemeryanz (changes requested):
This pull request now contains the changes from your other branch. I'll contact you on Slack to get this over the line.

andrewemeryanz:
> One more thing that would be super helpful for any users is to document ...

Issue created: https://github.com/anz-bank/sysl-go/issues/54

andrewemeryanz:
#major change.

andrewemeryanz (approved):

PR #53, opened 2020-03-30
Feat/handle error transform

tiagralee:
issue #1

Add a common.CustomError
Changed svc_handler.sysl to transform the error type to the CustomerError map
Changed svc_handler.sysl add a new function HandleError to map custom error, if not found return the common http error.
Update the tests for simple and dbendpoints.


codegen/tests/simple/servicehandler.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/anz-bank/sysl-go/validator"
 )
 
+// *BusinessLogicError error
+var BusinessLogicError = map[string]string{"name": "BusinessLogicError", "http_code": "1001", "http_message": "foo", "http_status": "500"}
```
andrewemeryanz: This needs to be:
```go
var BusinessLogicError CustomError = map[string]string{ ... }
```

so you can use it as an error directly:
```go
var e error = BusinessLogicError
```

codegen/tests/simple/simple_gencallback.go
```diff
@@ -41,3 +37,21 @@ func (c Callback) AddMiddleware(ctx context.Context, r chi.Router) {
 func (c Callback) BasePath() string {
 	return "/"
 }
+
+func (c CallbackWithMapError) MapError(ctx context.Context, cause error) *common.HTTPError {
```
andrewemeryanz: `MapError` should take a `ServerError`, not just an `error` to make this more explicit.

You can modify `CreateError` to return a `ServerError` as follows:

```go
func CreateError(ctx context.Context, kind Kind, message string, cause error) ServerError {
	// we may push the error to NR here
	if err := CheckContextTimeout(ctx, message, cause); err != nil {
		return *err
	}

	e, ok := cause.(*ServerError)
	if !ok {
		return ServerError{Kind: kind, Message: message, Cause: cause}
	}

	return *e
}

func CheckContextTimeout(ctx context.Context, message string, cause error) *ServerError {
	if ctx.Err() == context.DeadlineExceeded {
		return &ServerError{Kind: DownstreamTimeoutError, Message: message, Cause: cause}
	}
	return nil
}
```

common/error_handler.go
```diff
@@ -13,10 +14,30 @@ const (
 	unknownError          = "Unknown Error"
 )
 
-func HandleError(ctx context.Context, err error) HTTPError {
+func HandleError(ctx context.Context, w http.ResponseWriter, kind Kind, message string, cause error, httpErrorMapper func(context.Context, error) *HTTPError) {
+	err := CreateError(ctx, kind, message, cause)
 	logEntry := GetLogEntryFromContext(ctx)
-	logEntry.Error(err)
```
andrewemeryanz, line 18: Can we just log the `ServerError` here instead of the `cause`? The `ServerError` contains the `cause`.

andrewemeryanz (changes requested):
Couple of changes requested. Thanks.

codegen/tests/simple/simple_gencallback.go
tiagralee: This CreateError is referenced by other places as well those requires a generic error, do we need to create separate CreateServerError method instead?

andrewemeryanz: Ok. Leave it for now.

codegen/tests/simple/simple_gencallback.go
```diff
@@ -41,3 +37,21 @@ func (c Callback) AddMiddleware(ctx context.Context, r chi.Router) {
...
+	return &common.HTTPError{HTTPCode: 500, Code: "1001", Description: "foo"}
```
andrewemeryanz: This should be passed in from the test rather than hardcoded (to give your tests flexibility). Add a field to the `Callback` and remove `CallbackWithMapError`:
```go
type Callback struct{
    MapError *common.HTTPError
}

func (c CallbackWithMapError) MapError(ctx context.Context, cause error) *common.HTTPError {
    return c.MapError
}
```

then just pass in your error to test:
```go
cb := Callback{ &common.HTTPError{ ... } }
``

**Review by andrewemeryanz — APPROVED** [review]

File: codegen/transforms/svc_types.sysl [hunk 5]
```diff
@@ -309,7 +309,7 @@ CodeGenTransform:
 
       Comment = "// Code generated by sysl DO NOT EDIT.\n"
 
-      TopLevelDecl = unused | GoStructs(app.types) | GoSetType(app.types) | RequestTypes(app.endpoints, module) | validatorDecl(app.types) |  addDecl(app.types) | lookupDecl(app.types) | GoAlias(app.alias where(HasPrefix(.key, "EXTERNAL_") == false && .key != "Empty")) | GoEmptyType(app.alias where(.key == "Empty"))
+      TopLevelDecl = unused | GoStructs(app.types where(.attrs.patterns == null)) | GoSetType(app.types where(.attrs.patterns == null)) | RequestTypes(app.endpoints, module) | validatorDecl(app.types where(.attrs.patterns == null)) |  addDecl(app.types where(.attrs.patterns == null)) | lookupDecl(app.types where(.attrs.patterns == null)) | GoAlias(app.alias where(HasPrefix(.key, "EXTERNAL_") == false && .key != "Empty")) | GoEmptyType(app.alias where(.key == "Empty"))
```
gkanz: This will generate spurious warnings, should use `where("patterns" !in .attrs))` instead of `where(.attrs.patterns == null)`.

!in is missing in Sysl. I have the necessary changes: https://github.com/anz-bank/sysl/tree/add-not-in-binexpr but I need to write the tests.

common/error_handler.go
```diff
@@ -13,10 +14,26 @@ const (
...
+	err := CreateError(ctx, kind, message, cause)
 	logEntry := GetLogEntryFromContext(ctx)
 	logEntry.Error(err)
+	httpError := httpErrorMapper(ctx, err)
 
+	if httpError == nil {
+		switch t := cause.(type) {
+		case CustomError:
+			e := t.HTTPError(ctx)
+			httpError = &e
```
gkanz: Can this be a single line?

common/error_handler.go
```diff
@@ -13,10 +14,26 @@ const (
...
+		default:
+			e := MapError(ctx, err)
+			httpError = &e
```
gkanz, line 29: Can this be a single line?

common/custom_error.go
```diff
@@ -0,0 +1,34 @@
...
+	httpStatusString := getOrDefault(e, "http_status", "500")
+	httpStatus, err := strconv.Atoi(httpStatusString)
+	if err != nil {
+		logEntry := GetLogEntryFromContext(ctx)
+		logEntry.Error(fmt.Sprintf("invalid http_status: %s for: %s", httpStatusString, e["name"]))
+		httpStatus = 500 //nolint: // TODO: use constant for internal server error
+	}
+	httpCode := getOrDefault(e, "http_code", "")
+	httpMessage := getOrDefault(e, "http_message", "")
+	return HTTPError{httpStatus, httpCode, httpMessage}
```
gkanz: Is there a reason we are making a copy of the `HTTPError` here? Could `httpMessage` be large? Would returning a pointer help with:
```Go
e := t.HTTPError(ctx)
httpError = &e
```

common/error_handler.go
andrewemeryanz: > Can this be a single line?

No unfortunately. `&MapError(ctx, err)` is invalid Go.

common/error_handler.go
andrewemeryanz: > Can this be a single line?

Yes, if `t.HTTPError` returns a pointer.

common/custom_error.go
andrewemeryanz: >  Could httpMessage be large?

It's unlikely to be of any great length but yes, a pointer can be returned here.

codegen/transforms/svc_types.sysl
tiagralee: This is not merged yet, can I leave if for now and create an issue to change it after the sysl update?

gkanz: OK, but you will get complaints about the warnings.

andrewemeryanz: @gkanz Can we merge this in with a `TODO` and revisit after your changes?

core/callback.go
```diff
@@ -19,7 +18,7 @@ type RestGenCallback interface {
 	// This is returned from the status endpoint
 	Config() validator.Validator
 	// HandleError allows custom HTTP errors to be added to `sys-go` errors
-	HandleError(ctx context.Context, w http.ResponseWriter, kind common.Kind, message string, cause error)
+	MapError(ctx context.Context, cause error) *common.HTTPError
```
andrewemeryanz: Minor request. Can we rename `cause` to `err`. The actual error passed is a wrapped error, not the original cause.

core/callback.go
```diff
@@ -19,7 +18,7 @@ type RestGenCallback interface {
 	// This is returned from the status endpoint
 	Config() validator.Validator
 	// HandleError allows custom HTTP errors to be added to `sys-go` errors
```
andrewemeryanz: Also, can we update this documentation:
```
MapError maps an error to an HTTPError in instances where custom error mapping is 
required. Return nil to perform default error mapping; defined as: 
1. CustomError.HTTPError if the original error is a CustomError, otherwise
2. common.MapError
```

andrewemeryanz (approved):

PR #61, opened 2020-04-01
Use callback interface specified in sysl-go instead of generating it each time

gkanz:
Instead of generating the callback interface in `requestrouter.go` (which doesn't change based on generated types), use one specified directly in Go.

Refer to individual commit comments for rationale of the groupings.

Due to the way Go handles interfaces this should not be a breaking change.


codegen/tests/cardspb/cards_api.pb.go
```diff
@@ -6,12 +6,11 @@ package cardspb
 import (
 	context "context"
 	fmt "fmt"
-	math "math"
-
 	proto "github.com/golang/protobuf/proto"
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
+	math "math"
```
ChloePlanet: This file, `codegen/tests/cardspb/wallet_api.pb.go` and `codegen/tests/simplepb/simplegrpc.pb.go` shouldn't be changed after `goimports -w .` run. Might related to Makefile?

ChloePlanet (approved):
LGTM

codegen/tests/cardspb/cards.pb.go
```diff
@@ -5,9 +5,8 @@ package cardspb
 
 import (
 	fmt "fmt"
-	math "math"
-
 	proto "github.com/golang/protobuf/proto"
+	math "math"
```
camscale: i have always run `goimports` on the generated file, which is what gives the original ordering you had. I think it is better to do that. maybe.

camscale (approved):
LGTM modulo goimports comment.

This is looking much nicer. I like that it is easier to have one place to document the callback interfaces, which we can incrementally improve if any questions come in about it. Removing more lines than being added is another bonus.

:+1:

codegen/tests/cardspb/cards_api.pb.go
gkanz: Will fix. Thanks.

codegen/tests/cardspb/cards.pb.go
gkanz: Done.

PR #67, opened 2020-04-03
Populate model dependencies in make

gkanz:
Make was not noticing when models changed and so did not rebuild the outputs.

The issue was due to the target-specific variable `$(MODEL)` not being populated in prerequisites. Instead of using `.SECONDEXPANSION` to solve the problem, the pre-requisite is now specified as a normal variable.


ghost (approved):
👍 works for me

camscale (approved):

Issue #64, opened 2020-04-02
Logging issue in common lib


JackLeeHal:
1. https://github.com/anz-bank/sysl-go/blob/master/common/corerequestcontext.go#L100
not using the correct response writer
should add line:  `w = tl.RespWrapper`

2. not logging the incoming request and output response of bff round trip, just log the round trip of downstream.

JackLeeHal:
Fixed by https://github.com/anz-bank/sysl-go/pull/68

PR #68, opened 2020-04-03
fix logging issue

JackLeeHal:
Regarding issue:https://github.com/anz-bank/sysl-go/issues/64
1. Using the correct response writer
2. Factor out the req/resp debug logging so it can be used for incoming requests also

### DIFF
--- a/common/corerequestcontext.go
+++ b/common/corerequestcontext.go
@@ -95,9 +95,15 @@ func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler)
 			ctx = internal.AddResponseBodyMonitorToContext(ctx)
 			defer internal.CheckForUnclosedResponses(ctx, entry)
 
+			reqLogger, entry := internal.NewRequestLogger(entry, r)
+			w = reqLogger.ResponseWriter(w)
+			defer reqLogger.FlushLog()
+			LoggerToContext(ctx, logger, entry)
+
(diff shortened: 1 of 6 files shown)


JackLeeHal:
@anz-gordonj7 Can you help review on this? After move common lib, there are logging issues being reported.

common/corerequestcontext.go
```diff
@@ -95,9 +95,15 @@ func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler)
 			ctx = internal.AddResponseBodyMonitorToContext(ctx)
 			defer internal.CheckForUnclosedResponses(ctx, entry)
 
+			reqLogger, entry := internal.NewRequestLogger(entry, r)
+			w = reqLogger.ResponseWriter(w)
+			defer reqLogger.FlushLog()
+			LoggerToContext(ctx, logger, entry)
```
ghost: the return value of this function is the context with the logger, you need to be using this value

ghost (changes requested):

common/corerequestcontext.go
JackLeeHal: addressed

ghost (approved):

JackLeeHal:
@anz-gordonj7 @gkanz anyone can help merge this PR? I don't have the access.

PR #69, opened 2020-04-07
Attempt to handle bytes internal type correctly

ghost:
@gkanz need help finishing this off please. The generated code needs to be changed to pass the `[]byte` through from the generated types.

### DIFF
--- a/codegen/tests/simple/servicehandler.go
+++ b/codegen/tests/simple/servicehandler.go
@@ -22,6 +22,7 @@ var BusinessLogicError2 common.CustomError = map[string]string{"name": "Business
 // Handler interface for Simple
 type Handler interface {
 	GetApiDocsListHandler(w http.ResponseWriter, r *http.Request)
+	GetGetSomeBytesListHandler(w http.ResponseWriter, r *http.Request)
 	GetJustOkAndJustErrorListHandler(w http.ResponseWriter, r *http.Request)
 	GetJustReturnErrorListHandler(w http.ResponseWriter, r *http.Request)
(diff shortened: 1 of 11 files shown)


codegen/tests/simple/servicehandler.go
```diff
@@ -80,6 +81,38 @@ func (s *ServiceHandler) GetApiDocsListHandler(w http.ResponseWriter, r *http.Re
...
+
+	pdf, err := s.serviceInterface.GetGetSomeBytesList(ctx, &req, client)
+	if err != nil {
+		common.HandleError(ctx, w, common.DownstreamUnexpectedResponseError, "Downstream failure", err, s.genCallback.MapError)
+		return
+	}
+
+	headermap, httpstatus := common.RespHeaderAndStatusFromContext(ctx)
+	restlib.SetHeaders(w, headermap)
+	restlib.SendHTTPResponse(w, httpstatus, pdf)
```
ghost: @gkanz the `pdf` type is `type Pdf []byte` (and of course a pointer at this point). It is passed directly to `restlib.SendHTTPResponse` which attempts to cast it to a `[]byte` which fails. we need this code to pass the correct type down please

gkanz: I have updated the `svc_handler.sysl` and sent you the diff. Caveats:

- only currently works where the type is declared in the same scope as the app.
- it generates warnings because Sysl doesn’t short circuit && evaluations.

gkanz (approved):
Approved with the caveats on the solution as noted in the PR conversation.

PR #70, opened 2020-04-07
Move ReadConfig to LoadConfig and validate

andrewemeryanz:
The config.ReadConfig method reads configuration from file and
populates passed-in config structs. To ensure that the returned
configuration is valid the method now validates the loaded
configuration before returning.

To avoid ambiguity and provide only a single mechanism to load
configuration, this method has been renamed to LoadConfig.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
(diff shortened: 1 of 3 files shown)


andrewemeryanz:
Closed in favour of: https://github.com/anz-bank/sysl-go/pull/81

PR #81, opened 2020-04-28
Move ReadConfig to LoadConfig and validate

andrewemeryanz:
The config.ReadConfig method reads configuration from file and
populates passed-in config structs. To ensure that the returned
configuration is valid the method now validates the loaded
configuration before returning.

To avoid ambiguity and provide only a single mechanism to load
configuration, this method has been renamed to LoadConfig.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"io/ioutil"
 
+	"github.com/anz-bank/sysl-go/validator"
+
 	"github.com/mitchellh/mapstructure"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -40,5 +43,20 @@ func ReadConfig(cfgFile string, defaultConfig *DefaultConfig, customConfig inter
 		return err
 	}
 
-	return decoder.Decode(c)
+	err = decoder.Decode(c)
+	if err != nil {
+		return err
+	}
+
+	err = validator.Validate(defaultConfig)
```
springwiz, line 51: The change needs to be complemented with validation rule configuration on the config struct. Are the validation rules part of another PR?

andrewemeryanz: Not sure what your question is. Validation will fail if children are invalid, even if the parent isn't explicitly a `Validator`.

springwiz: Just saw that the validation rules are already annotated on the config structs i.e. validate:"nonnil"`, may be got included in another PR.

springwiz (approved):
LGTM

config/config.go
andrewemeryanz: Yes. Not sure of the history of the validation rules, but they're untouched in this change.

PR #83, opened 2020-04-28
Add more http statuses to success criteria.

### LINKED ISSUE #80: HTTP response code 204 No Content has been consider as errors from Sysl
Hi everyone,

Across this issue for one of our downstream API, they return a 204 No Content response code for an empty list. 

At the moment, Sysl-go has treated 204 as an error in restlib.go. here is the code snippet:
```
    // OK
    if httpResponse.StatusCode == http.StatusOK ||
        httpResponse.StatusCode == http.StatusCreated ||
        httpResponse.StatusCode == http.StatusAccepted {
        return unmarshal(httpResponse, respBody, okResponse)
    }

    // Error
    result, err := unmarshal(httpResponse, respBody, errorResponse)
    if err != nil {
        return nil, err
    }
```
[Link attached here](https://github.com/anz-bank/sysl-go/blob/0a505042aebe9debfa30f748a3a4a6ca8f57a279/restlib/restlib.go#L132)

Not sure whether this was intended, we are open to suggestions from you.

Thanks,
Ping

springwiz:
Add more http statuses in the success criteria, it includes only http.StatusOK, http.StatusCreated and
http.StatusAccepted for now.

1. Add more http statuses in the success criteria.
2. Fix code to send back an empty response, if the downstream returns a nil response.

Closes #80

### DIFF
--- a/restlib/restlib.go
+++ b/restlib/restlib.go
@@ -41,6 +41,11 @@ func unmarshal(resp *http.Response, body []byte, respStruct interface{}) (*HTTPR
 		panic("unmarshal expecting a non-nil http.Response")
 	}
 	if respStruct == nil || body == nil || len(body) == 0 {
+		// Obtain the respStruct's dynamic type and check if its a pointer
+		if p := reflect.TypeOf(respStruct); p != nil && p.Kind() == reflect.Ptr {
+			// Dereference the dynamic pointer type and pass the created zero value
+			return makeHTTPResult(resp, body, reflect.New(p.Elem()).Interface()), nil
+		}
 		return makeHTTPResult(resp, body, nil), nil
 	}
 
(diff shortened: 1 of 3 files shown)


restlib/restlib.go
```diff
@@ -129,9 +134,14 @@ func DoHTTPRequest(ctx context.Context, client *http.Client, method string,
...
-		httpResponse.StatusCode == http.StatusCreated ||
-		httpResponse.StatusCode == http.StatusAccepted {
+	switch httpResponse.StatusCode {
+	case http.StatusOK,
+		http.StatusCreated,
+		http.StatusAccepted,
+		http.StatusNonAuthoritativeInfo,
+		http.StatusNoContent,
+		http.StatusResetContent,
+		http.StatusPartialContent:
```
andrewemeryanz: I would remove `206 Partial Content` from the list. We never send a range header in the request and we always assume the full response is received.

andrewemeryanz (changes requested):
One minor change. Otherwise looks good.

restlib/restlib.go
springwiz: ok

andrewemeryanz (approved):
Good.

PR #88, opened 2020-04-30
Remove library config and admin server from Manager -- new PR

springwiz:
This change does the following:

Removes the library config from the Manager interface in favour
of passing it in on server startup.

Removes the admin server from the Manager interface in favour of
including it in the library config.

Renames HandlerInitialiser to RestHandlerInitialiser.

Fix broken tests.

Refactor GrpcGenCallback and Add config to it.

Add E2E test case for admin server.


andrewemeryanz:
Dormant.

Issue #66, opened 2020-04-03
Replace logrus with anz-bank/pkg logging


andrewemeryanz:
Reference: https://github.com/anz-bank/pkg/tree/master/log
Required for: https://github.com/anzx/vcards/issues/108

PR #91, opened 2020-05-04
Feature/use pkg logger

tiagralee:
https://github.com/anz-bank/sysl-go/issues/66
This change replace the logrus with pkg logger

Fix all related tests.


common/internal/roundtripper_test.go
```diff
@@ -128,10 +124,11 @@ func TestLoggingTransport_RoundTripLogFields(t *testing.T) {
 	respFound := false
 	statusFound := false
 	for _, entry := range hook.Entries {
-		require.Equal(t, "this is trace id", entry.Data[distributedTraceIDName])
-		require.Equal(t, "this is span id", entry.Data[distributedSpanIDName])
-		require.Nil(t, entry.Data[distributedParentSpanIDName])
-		if entry.Level == logrus.DebugLevel {
+		entryTraceIDName, _ := entry.Data.Get(distributedTraceIDName)
+		require.Equal(t, "this is trace id", entryTraceIDName)
+		// require.Equal(t, "this is span id", entry.Data.Get(distributedSpanIDName))
```
andrewemeryanz: Commented out code. Are these checks necessary?

core/server.go
```diff
@@ -144,9 +130,8 @@ func WithGrpcManager(manager GrpcManager) ServerOption {
 }
 
 //nolint:gocognit // Long method names are okay because only generated code will call this, not humans.
-func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, logger *logrus.Logger, promRegistry *prometheus.Registry) error {
+func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, promRegistry *prometheus.Registry) error {
```
andrewemeryanz: This function needs to remain backwards compatible, meaning `ServerParams` needs to continue to support `WithLogrusLogger`.

core/tls_log_filter.go
```diff
@@ -3,20 +3,19 @@ package core
...
 
 type TLSLogFilter struct {
-	logger *logrus.Logger
+	logger log.Logger
 	re     *regexp.Regexp
 }
 
 func (t *TLSLogFilter) Write(p []byte) (n int, err error) {
-	level := logrus.WarnLevel
 	if t.re.Match(p) {
```
andrewemeryanz, line 15: I think this should be:
```go
if ... {
    t.logger.Debug(string(p))
} else {
    t.logger.Info(string(p))
}
```

core/tls_log_filter_test.go
```diff
@@ -1,46 +1,32 @@
 package core
 
-import (
-	"fmt"
-	"log"
-	"regexp"
-	"testing"
+// func TestTLSLogFilter_Write(t *testing.T) {
```
andrewemeryanz: Is this code not required any more or is it still a work in progress? If it's not required then it should be removed.

common/internal/requestlogger_test.go
```diff
@@ -43,22 +39,20 @@ func TestNopLogger_FlushLog(t *testing.T) {
 	require.Empty(t, hook.Entries)
 
 	// and should be no-ops
-	rw := httptest.NewRecorder()
-	require.Equal(t, rw, l.ResponseWriter(rw))
+	//rw := httptest.NewRecorder()
+	//require.Equal(t, rw, l.ResponseWriter(rw))
```
andrewemeryanz: Commented out code. Are these checks necessary?

common/internal/requestlogger.go
```diff
@@ -80,10 +82,11 @@ func (r *nopLogger) LogResponse(_ *http.Response)
 func (r *nopLogger) ResponseWriter(base http.ResponseWriter) http.ResponseWriter { return base }
 func (r *nopLogger) FlushLog()                                                   {}
 
-func NewRequestLogger(entry *logrus.Entry, req *http.Request) (RequestLogger, *logrus.Entry) {
-	if entry.Logger.IsLevelEnabled(logrus.DebugLevel) {
+func NewRequestLogger(ctx context.Context, req *http.Request, isDebug bool) (RequestLogger, context.Context) {
+	//nolint: // TODO: store debug logging flag against context (outside of logger): Context#withValue
```
andrewemeryanz: > Add the following function to support the `pkg` logger

Refer to the comment with the above text to help inform you about this area. Basically you want to store the verbose state of the logger in the context and pull it out here. The reason for this approach is that, while it's possible to use the `requestLogger` for each request and just allow the `pkg` logger to decide whether or not to log a particular entry, I'd prefer to avoid reading the body twice and caching it in a byte array if we have no intention of doing anything with it.

common/corerequestcontext.go
```diff
@@ -5,26 +5,20 @@ import (
 	"net/http"
 
 	"github.com/anz-bank/sysl-go/common/internal"
-
 	"github.com/sirupsen/logrus"
-)
 
-func GetLogEntryFromContext(ctx context.Context) *logrus.Entry {
```
andrewemeryanz, line 12: I'd like these methods to be deprecated rather than removed and function as they did before. If consumers use a `logrus.Logger` then they should still be able to pull it out of the context.

When deprecating, add clear documentation about what should be done instead (ie. use the `pkg` logger).

core/server.go
```diff
@@ -107,18 +105,6 @@ func WithRestManager(manager Manager) ServerOption {
 	return &restManagerOption{manager}
 }
 
-type loggerOption struct {
```
andrewemeryanz, line 110: Add the following function to support the `pkg` logger:
```
func WithPkgLogger(configs ...Config) ServerOption { ... }
```

Internally we only want to use the `pkg` logger, so whether `WithLogrusLogger` or `WithPkgLogger` or neither are called we will set a `pkg` logger against the context. I'll detail how these should be treated separately:

#### WithPkgLogger

Set a `pkg` logger against the context with the provided `configs`.

There's an interesting piece of code in the `NewRequestLogger` function that only performs a certain action if the `logrus` log level was high enough. The `logrus` log level was effectively being used to describe the verbosity of `sysl-go` logging (which although similar, are two different concepts).

We want to retain the idea of `sysl-go` logging verbosity with the `pkg` logger. We could add another `ServerOption` to indicate this but I'd rather leave it out of the interface for now and use the log level of the `logrus` logger (as previous) and the `verbose` level of the `pkg` logger to indicate this value.

The complication is that the `pkg` logger doesn't expose whether or not it is logging in a `verbose` manner, however you can check whether or not one of the `configs` is the verbose config, so this will have to do for now.

To communicate the `sysl-go` logging verbosity throughout the rest of `sysl-go`, a value should be added to the context in the `ServerParams` and pulled out when necessary.

#### WithLogrusLogger

Set a `pkg` logger against the context and use the `AddHooks` config value to forward `pkg` logger entries to the provided `logrus` logger.

Also add the `logrus` logger to the config to ensure it works in the same way as it did before.

Deprecate all exported methods that deal with the `logrus` logger directly.

#### No logger set

Treat this the same as if `WithPkgLogger` was called with no `configs`.

andrewemeryanz (changes requested):
Getting there. Changes requested to keep the api as backwards compatible as possible and improve `pkg` logger integration.

common/internal/roundtripper_test.go
tiagralee: put those back now.

common/internal/requestlogger_test.go
tiagralee: This test `TestNopLogger_FlushLog` is being removed, as the logrus level is not valid, test `TestLogger_FlushLog` will test the hook entry.

core/server.go
tiagralee: Just want to make sure if we have the configs for `WithPkgLogger`, how do we pass in the configs if we still want backward compatible?

common/error_handler.go
```diff
@@ -16,8 +19,11 @@ const (
 
 func HandleError(ctx context.Context, w http.ResponseWriter, kind Kind, message string, cause error, httpErrorMapper func(context.Context, error) *HTTPError) {
 	err := CreateError(ctx, kind, message, cause)
-	logEntry := GetLogEntryFromContext(ctx)
-	logEntry.Error(err)
+	bytes, e := json.Marshal(err)
+	if e != nil {
+		panic(e)
```
andrewemeryanz: We definitely shouldn't be panicking here. Json marshalling seems like an unusual operation here. Is it included because it mimics behaviour similar to the existing (logrus) behaviour? Can we just log the error:
```go
log.Error(ctx, err)
```

core/server.go
```diff
@@ -119,6 +120,25 @@ func WithLogrusLogger(logger *logrus.Logger) ServerOption {
...
 
+type ctxOption struct {
+	ctx context.Context
+}
+
+func (o *ctxOption) apply(params *ServerParams) {
+	params.Ctx = o.ctx
+}
+
+func WithPkgLogger(configs ...log.Config) ServerOption {
```
andrewemeryanz, line 165: I've pushed a branch [here](https://github.com/andrewemeryanz/sysl-go/tree/feature/use-pkg-logger) that shows you how I think the `pkg` logger should be integrated.

core/server.go
```diff
@@ -146,6 +166,7 @@ func WithGrpcManager(manager GrpcManager) ServerOption {
 //nolint:gocognit // Long method names are okay because only generated code will call this, not humans.
 func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, logger *logrus.Logger, promRegistry *prometheus.Registry) error {
 	return NewServerParams(ctx, name,
+		WithPkgLogger(),
```
andrewemeryanz, line 197: > WithPkgLogger()

You don't need this line. It should all work without.

common/corerequestcontext.go
```diff
@@ -114,6 +97,16 @@ func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler)
 
 type coreRequestContextKey struct{}
 
+// Deprecated: github.com/anz-bank/pkg/log is replacing the logrus
+func GetLogEntryFromContext(ctx context.Context) *logrus.Entry {
```
andrewemeryanz: Please move these back to their previous position within the file. It make reviewing and conflicts easier to manage.

andrewemeryanz (changes requested):

common/error_handler.go
tiagralee: Yes, this is to mimic the logrus log hook for testing. `log.Error(ctx, err)` just get the empty hook entry on the test.

core/server.go
andrewemeryanz: Can we remove this line?

common/internal/unclosedresponsebodymonitor_test.go
```diff
@@ -29,18 +29,15 @@ func TestCheckForUnclosedResponses(t *testing.T) {
 
 	// test
 	require.Panics(t, func() {
-		CheckForUnclosedResponses(ctx, logger.WithField("test", "test"))
+		CheckForUnclosedResponses(ctx)
 	})
-
-	for _, entry := range hook.Entries {
```
andrewemeryanz, line 35: Can we add a simple check here? It doesn't need to be equivalent but at least we should verify that something is logged.

common/internal/test_util.go
```diff
@@ -0,0 +1,27 @@
...
+type TestHook struct {
+	Entries []log.LogEntry
+}
+
+func (h *TestHook) OnLogged(entry *log.LogEntry) error {
+	h.Entries = append(h.Entries, *entry)
+	return nil
+}
+
+func NewTestContextWithLoggerHook() (context.Context, *TestHook) {
```
andrewemeryanz: This is largely duplicated from `common.test_util.go`. Can we unify these?

common/error_handler.go
```diff
@@ -16,8 +19,8 @@ const (
 
 func HandleError(ctx context.Context, w http.ResponseWriter, kind Kind, message string, cause error, httpErrorMapper func(context.Context, error) *HTTPError) {
 	err := CreateError(ctx, kind, message, cause)
-	logEntry := GetLogEntryFromContext(ctx)
-	logEntry.Error(err)
+	bytes, _ := json.Marshal(err)
```
andrewemeryanz: Is json marshalling necessary? What does the standard logger do if you just pass an error?

common/internal/requestlogger.go
```diff
@@ -93,7 +92,15 @@ func NewRequestLogger(entry *logrus.Entry, req *http.Request) (RequestLogger, *l
 			req.Body = ioutil.NopCloser(bytes.NewBuffer(b))
 			l.req.body.Write(b)
 		}
-		return l, l.e
+		return l, l.ctx
+	}
+	return &nopLogger{}, ctx
+}
+
+func getDebugLogLevelCtx(ctx context.Context) *logconfig.IsVerboseLogging {
```
andrewemeryanz: Could this function instead be:
```go
func isVerboseLogging(ctx context.Context) bool
```

core/recoverer_test.go
```diff
@@ -28,6 +23,5 @@ func TestRecoverer(t *testing.T) {
 	if res != nil {
 		defer res.Body.Close()
 	}
-
-	require.True(t, strings.Contains(buffer.String(), "Panic: Test"))
+	require.Panics(t, nil, "Panic: Test")
```
andrewemeryanz: This doesn't look identical. The old test validated that something was logged. The new test panics because it's passed a `nil` function value.

andrewemeryanz (changes requested):
Almost there.

common/error_handler.go
tiagralee: The look hook will get empty message if we don't do json marshalling.
eg: in `simple_test.go` line 212, 223, 235

`require.Equal(t, "{\"Kind\":2,\"Message\":\"foo\",\"Cause\":null}", hook.LastEntry().Message)`

common/internal/test_util.go
tiagralee: Extract this to the testutil package, cannot keep this part in the common/test_util will get import cycle error.

common/error_handler.go
andrewemeryanz: The log entry should contain the error within its `Data` field. 

The real question about this refactor is: does the logger log out the error in a meaningful way? `log.Error(ctx, err)` should be sufficient to log the error. If you want to test what the logger actually logs then you can add the `SetOutput` config when you're configuring the logger.

andrewemeryanz (approved):
Looks good.

PR #97, opened 2020-05-08
Feat/server builder

### LINKED ISSUE #87: Provide builder approach for Server
The current mechanism to start the server is through the `Server` method:

```go
func Server(ctx context.Context, name string, libraryConfig *config.LibraryConfig,
    hl RestManager, grpcHl GrpcManager,
    logger *logrus.Logger, promRegistry *prometheus.Registry) error { ... }
```

To provide flexibility and avoid future breaking changes, a builder approach should be used:

```go
type ServerParams struct {
    Ctx                context.Context
    Name               string
    Config             *config.DefaultConfig
    restManager        RestManager
}

func NewServerParams(ctx context.Context, name string, config *config.DefaultConfig) *ServerParams {
    return &ServerParams{Ctx: ctx, Name: name, Config: config,}
}

func (params *ServerParams) Start(opts ...ServerOption) error {
    for _, o := range opts {
        if err := o.Apply(params); err != nil {
            return err
        }
    }
    ...
    return nil
}

type ServerOption interface {
    Apply(params *ServerParams) error
}

type restManagerOption struct {
    restManager RestManager
}

func (o *restManagerOption) Apply(params *ServerParams) error {
    params.restManager = o.restManager
    return nil
}

func WithRestManager(manager RestManager) ServerOption {
}
```

tiagralee:
Provide builder approach for Server #87
Use a builder approach to start the server.


core/server.go
```diff
@@ -5,12 +5,147 @@ package core
 import (
 	"context"
 
+	"github.com/anz-bank/sysl-go/config"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
+type ServerParams struct {
```
springwiz, line 13: Please add test cases for starting server with the various options.

andrewemeryanz:
Haven't looked extensively, but is the existing `Server` method updated to use the `ServerParams`?

tiagralee:
> Haven't looked extensively, but is the existing `Server` method updated to use the `ServerParams`?

No, didn't touch `Server` method

andrewemeryanz:
> No, didn't touch `Server` method

Ok. Please update the `Server` method to use `ServerParams` internally. We only want one underlying mechanism to start a server.

Issue #96, opened 2020-05-07
Transform failure: svc_types.sysl


andrewemeryanz:
A transform has failed with: `Failed to get key: type_ref`. The stacktrace contains information about the transform file but no information about the source file: `foo.sysl`. The source file is large so I'm not sure if the issue is with the source file or the transform.

### Version

`sysl-go v0.17.0`

### Stacktrace
```
sysl codegen --outdir internal/gen/foo --root-transform github.com/anz-bank/sysl-go/codegen --grammar grammars/go.gen.g --start goFile --basepath github.com/anzx/join-bff/internal/gen --app-name Foo --transform transforms/svc_types.sysl specs/backend/foo.sysl
WARN[0009] Failed to get key: type_ref. Map has following keys:  end="278:41" filename=transforms/svc_types.sysl start="278:40"
WARN[0009]      name                                         end="278:41" filename=transforms/svc_types.sysl start="278:40"
WARN[0009]      type                                         end="278:41" filename=transforms/svc_types.sysl start="278:40"
WARN[0009]      primitive                                    end="278:41" filename=transforms/svc_types.sysl start="278:40"
WARN[0009]      attrs                                        end="278:41" filename=transforms/svc_types.sysl start="278:40"
WARN[0009]      optional                                     end="278:41" filename=transforms/svc_types.sysl start="278:40"
ERRO[0009] Evaluation Failed:                           
ERRO[0009] ... Eval -> *sysl.Expr_Transform_             end="259:7" filename=transforms/svc_types.sysl start="252:6" text="let packages = Split(payload, \".\") -> <set of string>(p:\n        let depList = module.apps where(.value.name == p) -> <set of string> (dep:\n          pkg = getPackage(dep.value).pname\n          appName = p\n        )\n        pkg = depList flatten(.pkg)\n        appName = depList flatten(.appName)\n      )\n"
ERRO[0009] ... Eval -> *sysl.Expr_Transform_             end="262:5" filename=transforms/svc_types.sysl start="251:4"
ERRO[0009] ... Call -> getPackagePrefix()                end="278:57" filename=transforms/svc_types.sysl start="278:12" text="let pkg = getPackagePrefix(p.type_ref, module)"
[...]

PR #98, opened 2020-05-08
Fixes issue #96

ashwinsajiv:


andrewemeryanz:
The example file (passed out of band) turns a `sysl` definition like the following:
```
!type Foo:
    dataItems <: sequence of data?
...
!alias Data:
    string
```

into the following struct:
```go
type Foo struct {
    DataItems    []Data
}
```

Which results in a compilation error because `Data` is undefined. I suspect the generated code should be:
```go
type Foo struct {
    DataItems    []string
}
```

@springwiz Would this be related to this change or is it a separate issue?

springwiz:
> The example file (passed out of band) turns a `sysl` definition like the following:
> 
> ```
> !type Foo:
>     dataItems <: sequence of data?
> ...
> !alias Data:
>     string
> ```
> 
> into the following struct:
> 
> ```go
> type Foo struct {
>     DataItems    []Data
> }
> ```
> 
> Which results in a compilation error because `Data` is undefined. I suspect the generated code should be:
> 
> ```go
> type Foo struct {
>     DataItems    []string
> }
> ```
> 
> @springwiz Would this be related to this change or is it a separate issue?

There was a change done to support returning naked string parameters from the endpoints, and that is how !alias Data: string started getting used more. I am not sure, if it was intended to be used in this way. But using !alias Data: string should create a new go "type Data string". Thus compilation shouldn't fail anyways.

andrewemeryanz (approved):

andrewemeryanz:
> There was a change done to support returning naked string parameters from the endpoints, and that is how !alias Data: string started getting used more. I am not sure, if it was intended to be used in this way. But using !alias Data: string should create a new go "type Data string". Thus compilation shouldn't fail anyways.

Ok. So the issue I had is that my source `openapi3` file includes a `EXTERNAL_Foo` alias which is (correctly) being ignored by the transforms. So the behaviour is correct.

PR #101, opened 2020-05-12
Feat/server builder

### LINKED ISSUE #87: Provide builder approach for Server

tiagralee:
Provide builder approach for Server #87

### DIFF
--- a/core/server.go
+++ b/core/server.go
@@ -9,17 +9,35 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+type ServerParams struct {
+	Ctx                context.Context
+	Name               string
+	logger             *logrus.Logger
+	restManager        Manager
+	grpcManager        GrpcManager
+	prometheusRegistry *prometheus.Registry
+}
+
 //nolint:gocognit // Long method names are okay because only generated code will call this, not humans.
-func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, logger *logrus.Logger, promRegistry *prometheus.Registry) error {
-	mWare := prepareMiddleware(name, logger, promRegistry)
+func NewServerParams(ctx context.Context, name string, opts ...ServerOption) *ServerParams {
+	params := &ServerParams{Ctx: ctx, Name: name}
+	for _, o := range opts {
+		o.apply(params)
+	}
+	return params
+}
+
+//nolint:gocognit // Long method are okay because only generated code will call this, not humans.
+func (params *ServerParams) Start() error {
+	mWare := prepareMiddleware(params.Name, params.logger, params.prometheusRegistry)
 
 	var restIsRunning, grpcIsRunning bool
 
 	// Run the REST server
 	var listenAdmin func() error
-	if hl != nil && hl.AdminServerConfig() != nil {
+	if params.restManager != nil && params.restManager.AdminServerConfig() != nil {
 		var err error
-		listenAdmin, err = configureAdminServerListener(hl, logger, promRegistry, mWare.admin)
+		listenAdmin, err = configureAdminServerListener(params.restManager, params.logger, params.prometheusRegistry, mWare.admin)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


core/server.go
```diff
@@ -5,21 +5,102 @@ package core
...
+	Ctx                context.Context
+	Name               string
+	Config             *config.DefaultConfig
+	logger             *logrus.Logger
+	restManager        RestManager
+	grpcManager        GrpcManager
+	prometheusRegistry *prometheus.Registry
+}
+
+func NewServerParams(ctx context.Context, name string, config *config.DefaultConfig) *ServerParams {
```
andrewemeryanz: Can the `NewServerParams` function signature become:
```go
func NewServerParams(ctx context.Context, name string, opts ...ServerOption) error
```

and the `Start` function become:
```go
func (params *ServerParams) Start() error
```

The `config.DefaultConfig` isn't required to start the server.

core/server.go
```diff
@@ -5,21 +5,102 @@ package core
...
+type loggerOption struct {
+	logger *logrus.Logger
+}
+
+func (o *loggerOption) Apply(params *ServerParams) error {
+	params.logger = o.logger
+	return nil
+}
+
+func WithLogger(logger *logrus.Logger) ServerOption {
```
andrewemeryanz: Rename to `WithLogrusLogger`

core/server.go
```diff
@@ -5,21 +5,102 @@ package core
...
+}
+
+func WithGrpcManager(manager GrpcManager) ServerOption {
+	return &grpcManagerOption{manager}
+}
+
 //nolint:gocognit // Long method names are okay because only generated code will call this, not humans.
-func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, logger *logrus.Logger, promRegistry *prometheus.Registry) error {
-	mWare := prepareMiddleware(name, logger, promRegistry)
+func Server(params *ServerParams) error {
```
andrewemeryanz: I'd like this function signature to stay as it was to keep backwards compatibility. Internally it should route functionality to `ServerParams`:

```go
func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, logger *logrus.Logger, promRegistry *prometheus.Registry) error {
    return NewServerParams(ctx, name,
            WithLogrusLogger(logrus),
            WithRestManager(hl),
            WithGrpcManager(grpcHl),
            WithPrometheusRegistry(promRegistry))
        .Start()
```

andrewemeryanz (changes requested):
Couple of structural changes requested. General approach good.

core/server.go
```diff
@@ -5,21 +5,43 @@ package core
...
 
+	"github.com/anz-bank/sysl-go/config"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 )
 
+type ServerParams struct {
+	Ctx                context.Context
+	Name               string
+	Config             *config.DefaultConfig
```
andrewemeryanz: `Config` can be removed.

core/server.go
```diff
@@ -72,3 +94,68 @@ func Server(ctx context.Context, name string, hl Manager, grpcHl GrpcManager, lo
 
 	return <-errChan
 }
+
+type ServerOption interface {
+	Apply(params *ServerParams) error
```
andrewemeryanz: Change function signature to `apply(params *ServerParams)` so that:
- we're not exporting the function, and 
- error checking can be done in `Start`

andrewemeryanz (changes requested):
Minor changes requested.

andrewemeryanz (approved):
Looks good.

PR #102, opened 2020-05-12
Generate NewDefaultConfig method with downstream config.

### LINKED ISSUE #85: Generate NewDefaultConfig method with downstream config
The `GenCodeConfig` struct within `sysl-go` uses the `interface{}` type for the `Downstream` property:
```go
type GenCodeConfig struct {
    Upstream   UpstreamConfig `yaml:"upstream"`
    Downstream interface{}    `yaml:"downstream"`
}
```

This is required because the `Downstream` struct is generated:
```go
type DownstreamConfig struct {
	ContextTimeout time.Duration `yaml:"contextTimeout"`
}
```

To simplify generation, the `NewDefaultConfig` method should be created that injects the generated value:
```go
func NewDefaultConfig() config.DefaultConfig {
    return config.DefaultConfig{
        Library: config.LibraryConfig{},
        GenCode: config.GenCodeConfig{
            Downstream: &foo.DownstreamConfig{}, // generated type
        },
    }
}
```

springwiz:
Generate NewDefaultConfig method with downstream config.

1. Change transforms to generate NewDefaultConfig.
2. Fix test cases.

Fixes #85


andrewemeryanz (approved):
Looks good.

PR #112, opened 2020-05-27
Fix downstream to handle return <: ok.

### LINKED ISSUE #104: svc_interface.sysl does not handle untyped `return ok` response
given an endpoint like this:
```
foo:
    /orders:
        POST:
            return error <: Error
            return ok
```

the svc_client.sysl transform will create a func `PostOrders(ctx context.Context, req *PostOrdersRequest) (*http.Header, error)` however the svc_interface.sysl does not handle the untyped ok and it generates `PostOrders(ctx context.Context, req *PostOrdersRequest) error` instead. This ends up causing a compile time failure trying to initialisae the service interface.

springwiz:
Allows downstream to handle return <: ok.

1. Fixes the svc_interface transform.

Closes #104

### DIFF
--- a/codegen/transforms/svc_interface.sysl
+++ b/codegen/transforms/svc_interface.sysl
@@ -93,7 +93,10 @@ CodeGenTransform:
         let tn = ep.value.ret where (.key != "error" && .value != "ok" && .value != "error") -> <sequence of string>(t:
           out = makeType(t.value, package).out
         )
-        TypeName = Join(tn flatten(.out) | ["error"], ", ")
+        let ok = ep.value.ret where (.value == "ok") -> <sequence of string>(t:
+          out = "*http.Header"
+        )
+        TypeName = Join(tn flatten(.out) | ok flatten(.out) | ["error"], ", ")
       )
     )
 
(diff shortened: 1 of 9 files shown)


codegen/transforms/svc_interface.sysl
```diff
@@ -93,7 +93,10 @@ CodeGenTransform:
         let tn = ep.value.ret where (.key != "error" && .value != "ok" && .value != "error") -> <sequence of string>(t:
           out = makeType(t.value, package).out
         )
-        TypeName = Join(tn flatten(.out) | ["error"], ", ")
+        let ok = ep.value.ret where (.value == "ok") -> <sequence of string>(t:
```
andrewemeryanz, line 96: I have a feeling that headers are returned in another situation. Is this correct?

Returning headers for an unknown type seems a little arbitrary. I would have assumed that the successful result type would be either `*http.Response` (so the caller is free to inspect the entire response) or `interface{}` (to simply indicate a successful response).

springwiz: This just fixes a part that should have been taken care when we started accommodating for return <: ok. The intent is to surely reengineer return <: ok support but not necessarily as part of this bug fix.

andrewemeryanz: I see. So the previous intention (right or wrong) was that an untyped return would result in an `*http.Headers` result, and this pull request just fixes a related issue. This is fine.

andrewemeryanz (approved):

codegen/transforms/svc_interface.sysl
ghost: @andrewemeryanz yes, returning the full restlib.HTTPResponse would be much better

PR #114, opened 2020-05-27
include endpoint params in method name #110

### LINKED ISSUE #110: <critical> path var in endpoint does not get added to the client interface name, causes duplicate func names
sysl:
```
/orders/states:
        GET ?q=string?:
            | No description.
            return error
            return ok
/orders/{id<:int64}/states:
        GET ?q=string?&limit=int32?&offset=int32?:
            | No description.
            return error
            return ok <: OrderStateDto
```

Both of these endpoints generates a func named `GetOrdersStatesList(ctx context.Context, req *GetOrdersStatesListRequest) `, and both have the same named req type (which is also generated twice, once for each with different fields).

I think the required fix is that the second endpoint should take the path var names into account when naming, so something like `GetOrdersIdStatesList(...)` (which yeah, looks ugly but will have to do)

tiagralee:
Generate method name to include route params

Add an example in simple.sysl
Fixes the svc_handler transform.
Fixes the svc_interface transform.
Fixes the svc_client transform.
Fixes the svc_type transform.
Fixes the svc_router transform.

Closes #110

### DIFF
--- a/codegen/testdata/simple/simple.sysl
+++ b/codegen/testdata/simple/simple.sysl
@@ -19,9 +19,13 @@ Simple "Simple Server" [package="simple"]:
         GET:
             return ok <: str
 
-    /raw/{id<:string}/states:
+    /raw/{id<:string}/states [~vars_in_url_name]:
         GET:
-            return ok <: str    
+            return ok <: str
+
+    /raw/{id<:string}/states2:
+        GET:
+            return ok <: str     
 
     /raw-int:
         GET:
--- a/codegen/testdata/simple/simple.sysl
+++ b/codegen/testdata/simple/simple.sysl
@@ -23,7 +23,7 @@ Simple "Simple Server" [package="simple"]:
         GET:
             return ok <: str
 
-    /raw/{id<:string}/states2:
+    /raw/{id<:int64}/states2:
         GET:
             return ok <: str     
 
--- a/codegen/testdata/simple/simple.sysl
+++ b/codegen/testdata/simple/simple.sysl
@@ -15,6 +15,14 @@ Simple "Simple Server" [package="simple"]:
         GET:
             return ok <: str
 
+    /raw/states:
+        GET:
+            return ok <: str
+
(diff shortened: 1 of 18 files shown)


ghost (approved):

ghost:
Looks good, but its a breaking change so beware :) would be nice if it was only added if required, or toggled on, but thanks!

anzdaddy (dismissed):
This needs to be opt-in. We shouldn't be noising up every API just for the benefit of badly designed ones.

tiagralee:
> This needs to be opt-in. We shouldn't be noising up every API just for the benefit of badly designed ones.

Thanks @anzdaddy , where can we add a toggle?

codegen/testdata/simple/simple.sysl
```diff
@@ -15,6 +15,18 @@ Simple "Simple Server" [package="simple"]:
         GET:
             return ok <: str
 
+    /raw/states:
+        GET:
+            return ok <: str
+
+    /raw/{id<:string}/states [~vars_in_url_name]:
```
tiagralee, line 22: @anz-gordonj7 Please note here, that the params type should be string for now,  int64 the generated code will get type mismatch error as the GetURLParam from the server-lib just returns a string. Let me know if string didn't works for you, I'll work on the conditions in transforms.

springwiz (changes requested):
Looks good overall, you could add a quick test case in the simple_test.go.

codegen/testdata/simple/simple.sysl
ghost: Thanks, but yeah, we need it to support int64 also please

tiagralee: No problems, I'll working on this.

springwiz (approved):
lgtm

ghost (approved):
Works for us thanks!

andrewemeryanz (approved):

PR #116, opened 2020-05-29
Added interceptor callback to grpcmanager

nydrani:
Added a hook to allow external clients to add interceptors for grpc server generation


andrewemeryanz (approved):
Looks good. Let me know when you've merged and I'll update [this](https://github.com/anz-bank/sysl-go/wiki/v0-Breaking-Changes) page with the breaking change.

PR #119, opened 2020-06-03
Add a way for user code to add additional fields to the common error errors

ghost:
Problem:

The error json object one of our endpoints wants requires adding an extra field to the top level json object.. i.e the `Location` field in:
`{"status":{"Location":"createOrder","code":"1013","description":"Downstream system is unavailable"}}`

This patch adds a general way to add new fields to that status object. The reason we opted to add the field here instead of using a custom object is to continue to use a common error handling system with sysl-go and internally.

User code then looks like:
```

func withLocation(err error, loc string) error {
	return common.WrappedError(err, common.KV{"Location", loc})
}

....

orderResp, err := n.createOrder(reqCtx, req, client)
	if err != nil {
		return nil, withLocation(err, "createOrder")
	}

```

### DIFF
--- a/common/http_error_response_writer.go
+++ b/common/http_error_response_writer.go
@@ -10,16 +10,52 @@ type HTTPError struct {
 	HTTPCode    int    `json:"-"`
 	Code        string `json:"code,omitempty"`
 	Description string `json:"description,omitempty"`
+
+	extraFields map[string]interface{}
+}
+
+func (httpError *HTTPError) AddField(key string, val interface{}) {
+	if httpError.extraFields == nil {
+		httpError.extraFields = map[string]interface{}{}
+	}
+	httpError.extraFields[key] = val
+}
+
+type KV struct {
+	K string
+	V interface{}
+}
+type wrappedError struct {
+	e      error
+	fields []KV
+}
+
+func (w wrappedError) Error() string {
+	return w.e.Error()
+}
+func WrappedError(err error, fields ...KV) error {
+	return wrappedError{
+		e:      err,
+		fields: fields,
+	}
 }
 
 type httpErrorResponse struct {
-	Status *HTTPError `json:"status"`
+	Status interface{} `json:"status"`
 }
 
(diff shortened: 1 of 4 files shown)


AriehSchneier (approved):

common/http_error_response_writer.go
```diff
@@ -10,16 +10,52 @@ type HTTPError struct {
...
+		fields: fields,
+	}
 }
 
 type httpErrorResponse struct {
-	Status *HTTPError `json:"status"`
+	Status interface{} `json:"status"`
 }
 
 func (httpError *HTTPError) WriteError(ctx context.Context, w http.ResponseWriter) {
```
AriehSchneier, line 49: Just thinking, what if we made this an interface method rather than an object method, then the additional fields below can come from another interface, then we don't need to touch the HTTPError struct itself?

common/http_error_response_writer.go
```diff
@@ -10,16 +10,52 @@ type HTTPError struct {
...
+	K string
+	V interface{}
+}
+type wrappedError struct {
+	e      error
+	fields []KV
+}
+
+func (w wrappedError) Error() string {
+	return w.e.Error()
```
AriehSchneier: Wouldn't you want to append the extra fields here too?

ghost: yeah, should hav done that, willl fix

common/http_error_response_writer.go
andrewemeryanz: I like this suggestion. Something like:
```go
type WritableHTTPResponse interface {  // new type introduced
    Write(ctx context.Context, w http.ResponseWriter)
}

type RestGenCallback interface {
    MapError(ctx context.Context, err error) *common.WritableHTTPResponse  // method modified to suit
}
```

@gkanz Do you have any thoughts?

ghost: That would be a breaking change for other users though.

PR #120, opened 2020-06-09
Add a comment to the servicehandler interface if an endpoint has a feature attribute

ghost:
I'm running some post-processing on the generated code and I need to be able to group endpoints by features (to further generate more code). This patch adds a short comment to the `Handler` interface on methods with this tag..

e.g in the sysl:
```
    /v2/nudges [feature="nudges"]:
```

generates:
``` 	 
GetV2NudgesListHandler ( w http.ResponseWriter , r *http.Request ) // ~feature: nudges 
```

Which should allow me to read the go AST and find the ~feature comment that I need.

### DIFF
--- a/codegen/grammars/go.gen.g
+++ b/codegen/grammars/go.gen.g
@@ -7,7 +7,7 @@ NamedImport: Name Import;
 TopLevelDecl: Comment '\n' (Declaration | FunctionDecl | MethodDecl);
 Declaration: VarDecl | VarDeclWithVal | ConstDecl | StructType | InterfaceType | AliasDecl;
 StructType : 'type' StructName 'struct' '{\n' FieldDecl* '}\n\n';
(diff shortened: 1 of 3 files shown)


andrewemeryanz (changes requested):
I can't see us adopting this proposal as it stands. While this workaround might advance a solution in your instance, a `feature` isn't meaningful in this context and not something we'd want to support long term.

A reasonably generic approach would be for handler methods to include endpoint descriptions:
```
/v2/nudges:
  GET:
    | ~feature: nudges
    ...
```

But if the purpose of adding the comment is to do additional code generation then I think you're better off generating it from the sysl specification, leveraging the new arrai transforms.

ghost:
Figured out how to do this from the arr.ai transform so yeah killing proposal

PR #124, opened 2020-06-16
update pkg and add set log caller

### LINKED ISSUE #66: Replace logrus with anz-bank/pkg logging

tiagralee:
This is a quick fix in addition to the pkg logger, to add the condition to set the log caller if logrus set the `reportCaller` to true.


core/server.go
```diff
@@ -58,6 +58,10 @@ func (params *ServerParams) Start() error {
 		}
 	}
 
+	if params.logrusLogger != nil && params.logrusLogger.ReportCaller {
+		ctx = log.WithConfigs(log.SetLogCaller(true)).Onto(ctx)
```
andrewemeryanz: I think this would be slightly cleaner inside the section above where you're already dealing with the `logrus` logger:
```go
configs = append(configs, log.SetLogCaller(params.logrusLogger.ReportCaller))
```

andrewemeryanz (changes requested):
Slight change request. Otherwise good.

andrewemeryanz (approved):
Looks good.

PR #126, opened 2020-06-16
insert logger to ctx in middleware #125

### LINKED ISSUE #125: CoreRequestContextMiddleware needs to insert the internal logger into the context
custom code might use `common.CoreRequestContextMiddleware(logger)` directly and want to pass in the logger to the context, the current implementation is `common.CoreRequestContextMiddleware()`  and we inject the logger to the context in serverparams
Suggested change make `CoreRequestContextMiddleware` to take the `*logrus.Logger` and insert to the context if it is not nil and no existing logger in the context

tiagralee:
insert logger to ctx in corecontextmiddleware #125
- update corecontextmiddleware to take in the *logrus logger params, insert into context if there is no logger in the context.

### DIFF
--- a/common/corerequestcontext.go
+++ b/common/corerequestcontext.go
@@ -88,10 +88,18 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
 	return nil
 }
 
-func CoreRequestContextMiddleware() func(next http.Handler) http.Handler {
+func CoreRequestContextMiddleware(logger ...*logrus.Logger) func(next http.Handler) http.Handler {
+	var ctxlogger *logrus.Logger
+	if len(logger) == 1 {
+		ctxlogger = logger[0]
+	}
 	return func(next http.Handler) http.Handler {
 		fn := func(w http.ResponseWriter, r *http.Request) {
-			ctx := log.With(traceIDLogField, GetTraceIDFromContext(r.Context())).Onto(r.Context())
+			ctx := r.Context()
+			if ctxlogger != nil {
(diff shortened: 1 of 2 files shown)


common/corerequestcontext.go
```diff
@@ -88,10 +88,15 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
 	return nil
 }
 
-func CoreRequestContextMiddleware() func(next http.Handler) http.Handler {
+func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler) http.Handler {
```
ghost: you could make it `logger... *logrus.Logger` if you want it as an optional param which would be fine.

common/corerequestcontext.go
```diff
@@ -88,10 +88,15 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
...
 	return func(next http.Handler) http.Handler {
 		fn := func(w http.ResponseWriter, r *http.Request) {
-			ctx := log.With(traceIDLogField, GetTraceIDFromContext(r.Context())).Onto(r.Context())
+			ctx := r.Context()
+			// insert the logger to the context if the logger is passed in and no existing logger exists
+			if logger != nil && getCoreContext(ctx).logger != nil {
+				ctx = LoggerToContext(ctx, logger, nil)
+			}
```
ghost: This could simply be done once before this func setup as its either always going to be needed or never

ghost (changes requested):

ghost (approved):

PR #130, opened 2020-06-18
refactor corerequestcontext to pass logger #129

### LINKED ISSUE #129: Logger not passed to request context
The `CoreRequestContextMiddleware` function was previously passed a `logrus` logger directly. This logger was then injected into the request context.

The function signature is currently setup to support a `logrus` logger:
```go
func CoreRequestContextMiddleware(logger ...*logrus.Logger) func(next http.Handler) http.Handler
```

The function needs to support both `pkg` and `logrus` loggers (and potentially other values in the future). Design consideration:

```go
func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler) http.Handler {
    ctx := common.LoggerToContext(context.Background(), logger, nil)
    return CoreRequestContextMiddlewareWithContext(ctx)
}

func CoreRequestContextMiddlewareWithContext(ctx context.Context) func(next http.Handler) http.Handler {
    // add loggers to request context
}
```

tiagralee:
- Update `CoreRequestContextMiddleware` in `corerequestcontext.go` to pass in the logger directly or from context.
- Update related tests

### DIFF
--- a/common/corerequestcontext.go
+++ b/common/corerequestcontext.go
@@ -12,12 +12,20 @@ import (
 
 // Deprecated: Use ServerParams.WithPkgLogger instead
 func GetLogEntryFromContext(ctx context.Context) *logrus.Entry {
-	return getCoreContext(ctx).entry
+	core := getCoreContext(ctx)
+	if core == nil {
+		return nil
+	}
+	return core.entry
 }
 
 // Deprecated: Use ServerParams.WithPkgLogger instead
 func GetLoggerFromContext(ctx context.Context) *logrus.Logger {
-	return getCoreContext(ctx).logger
+	core := getCoreContext(ctx)
+	if core == nil {
+		return nil
+	}
+	return core.logger
 }
 
 func NewLoggingRoundTripper(name string, base http.RoundTripper) http.RoundTripper {
@@ -88,17 +96,19 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
 	return nil
 }
 
-func CoreRequestContextMiddleware(logger ...*logrus.Logger) func(next http.Handler) http.Handler {
-	var ctxlogger *logrus.Logger
-	if len(logger) == 1 {
-		ctxlogger = logger[0]
-	}
+func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler) http.Handler {
+	ctx := LoggerToContext(context.Background(), logger, nil)
+	return CoreRequestContextMiddlewareWithContext(ctx)
+}
+
+func CoreRequestContextMiddlewareWithContext(ctx context.Context) func(next http.Handler) http.Handler {
+	ctxlogger := GetLoggerFromContext(ctx)
+	pkgLogger := log.From(ctx)
 	return func(next http.Handler) http.Handler {
 		fn := func(w http.ResponseWriter, r *http.Request) {
 			ctx := r.Context()
-			if ctxlogger != nil {
-				ctx = LoggerToContext(ctx, ctxlogger, nil)
-			}
+			ctx = LoggerToContext(ctx, ctxlogger, nil)
+			ctx = log.WithLogger(pkgLogger).Onto(ctx)
 			ctx = log.With(traceIDLogField, GetTraceIDFromContext(ctx)).Onto(ctx)
 
 			ctx = internal.AddResponseBodyMonitorToContext(ctx)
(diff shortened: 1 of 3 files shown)


common/corerequestcontext.go
```diff
@@ -88,11 +88,13 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
...
-	if len(logger) == 1 {
-		ctxlogger = logger[0]
-	}
+func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler) http.Handler {
+	ctx := LoggerToContext(context.Background(), logger, nil)
+	return CoreRequestContextMiddlewareWithContext(ctx)
+}
+
+func CoreRequestContextMiddlewareWithContext(ctx context.Context) func(next http.Handler) http.Handler {
+	ctxlogger := GetCoreContext(ctx).Logger
```
andrewemeryanz: Both `logrus` logger and `pkg` logger need to be added to the request context.

common/corerequestcontext.go
```diff
@@ -12,21 +12,21 @@ import (
...
 }
 
 func NewLoggingRoundTripper(name string, base http.RoundTripper) http.RoundTripper {
 	// temporary pass-through to get the real round tripper from the request context
 	return &tempRoundtripper{name, base}
 }
 
-type coreRequestContext struct {
-	logger *logrus.Logger
+type CoreRequestContext struct {
```
andrewemeryanz: I think if you just pass the context through then you don't need to export this?

core/http.go
```diff
@@ -138,7 +139,12 @@ func prepareMiddleware(ctx context.Context, name string, promRegistry *prometheu
 	result.addToBoth(Recoverer(ctx))
 
 	result.public = append(result.public, common.TraceabilityMiddleware(ctx))
-	result.addToBoth(common.CoreRequestContextMiddleware())
```
andrewemeryanz, line 141: Just pass the context through here and let the middleware pull out the bits it wants:
```go
result.addToBoth(common.CoreRequestContextMiddleware(ctx))
```

core/http_test.go
```diff
@@ -46,7 +46,7 @@ func Test_prepareMiddleware(t *testing.T) {
 			want: []func(http.Handler) http.Handler{
 				Recoverer(ctx),
 				common.TraceabilityMiddleware(ctx),
-				common.CoreRequestContextMiddleware(),
+				common.CoreRequestContextMiddleware(nil),
```
andrewemeryanz: Pass the context through here and let the middleware pull out the bits it wants:
```go
result.addToBoth(common.CoreRequestContextMiddleware(ctx))
```

andrewemeryanz (changes requested):

common/corerequestcontext.go
```diff
@@ -88,17 +88,24 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
...
+		ctxlogger = getCoreContext(ctx).logger
 	}
 	return func(next http.Handler) http.Handler {
 		fn := func(w http.ResponseWriter, r *http.Request) {
 			ctx := r.Context()
 			if ctxlogger != nil {
 				ctx = LoggerToContext(ctx, ctxlogger, nil)
 			}
+			logger := log.NewStandardLogger()
+			ctx = log.WithLogger(logger).Onto(ctx)
```
tiagralee: @andrewemeryanz Can't get pkg logger from the ctx passed in (no key to log.WithLogger) , so I'm using this approach. Not sure if this is the right way?

common/corerequestcontext.go
```diff
@@ -88,17 +88,24 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
...
+func CoreRequestContextMiddleware(logger *logrus.Logger) func(next http.Handler) http.Handler {
+	ctx := LoggerToContext(context.Background(), logger, nil)
+	return CoreRequestContextMiddlewareWithContext(ctx)
+}
+
+func CoreRequestContextMiddlewareWithContext(ctx context.Context) func(next http.Handler) http.Handler {
 	var ctxlogger *logrus.Logger
-	if len(logger) == 1 {
-		ctxlogger = logger[0]
+	if getCoreContext(ctx) != nil {
```
andrewemeryanz: I'd modify `common.GetLoggerFromContext` (and `common.GetLogEntryFromContext`) to return `nil` if no logger is stored, then just call this method directly from here.

common/corerequestcontext.go
andrewemeryanz: Use `log.From(ctx)` to retrieve the logger. Cache the logger outside of the `return func` to avoid it being executed on each call.

common/corerequestcontext.go
```diff
@@ -88,17 +88,24 @@ func UpdateResponseStatus(ctx context.Context, status int) error {
...
+		ctxlogger = getCoreContext(ctx).logger
 	}
 	return func(next http.Handler) http.Handler {
 		fn := func(w http.ResponseWriter, r *http.Request) {
 			ctx := r.Context()
 			if ctxlogger != nil {
```
andrewemeryanz: You can avoid this if condition here. Setting a `nil` logger is fine.

common/corerequestcontext.go
```diff
@@ -17,6 +17,9 @@ func GetLogEntryFromContext(ctx context.Context) *logrus.Entry {
 
 // Deprecated: Use ServerParams.WithPkgLogger instead
 func GetLoggerFromContext(ctx context.Context) *logrus.Logger {
+	if getCoreContext(ctx) == nil {
```
andrewemeryanz: As a general approach, cache the result of a function rather than executing the function twice. It will:
1. Improve runtime performance.
2. Avoid issues if function may return a different value on a subsequent call.
```go
core := getCoreContext(ctx)
if core == nil {
    return nil
}
return core.logger
```

common/corerequestcontext.go
```diff
@@ -17,6 +17,9 @@ func GetLogEntryFromContext(ctx context.Context) *logrus.Entry {
 
 // Deprecated: Use ServerParams.WithPkgLogger instead
 func GetLoggerFromContext(ctx context.Context) *logrus.Logger {
```
andrewemeryanz, line 23: Can you make this same change to `GetLogEntryFromContext`.

common/corerequestcontext.go
```diff
@@ -122,6 +127,9 @@ func CoreRequestContextMiddleware(logger ...*logrus.Logger) func(next http.Handl
 type coreRequestContextKey struct{}
 
 func getCoreContext(ctx context.Context) *coreRequestContext {
+	if ctx.Value(coreRequestContextKey{}) == nil {
```
andrewemeryanz: Same principle here as above. Cache then check.

andrewemeryanz (approved):
Great.

PR #131, opened 2020-06-21
add empty writter to logrus

### LINKED ISSUE #132: Logrus and pkg logger log twice
If the customer pass in a logrus to the service params, the logrus and pkg logger both will log duplicated entries.
The solution will be set an empty io writter against pkg logger, pkg logger just becomes a proxy that forwards all logs to logrus.

tiagralee:
set an empty io writter against pkg logger
pkg logger just becomes a proxy that forwards all logs to logrus

### DIFF
--- a/core/server.go
+++ b/core/server.go
@@ -21,6 +21,12 @@ type ServerParams struct {
 	grpcManager        GrpcManager
 	prometheusRegistry *prometheus.Registry
 }
+type emptyWriter struct {
+}
+
+func (e *emptyWriter) Write(p []byte) (n int, err error) {
+	return len(p), nil
+}
 
 //nolint:gocognit // Long method names are okay because only generated code will call this, not humans.
 func NewServerParams(ctx context.Context, name string, opts ...ServerOption) *ServerParams {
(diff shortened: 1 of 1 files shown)


common/empty_writter.go
```diff
@@ -0,0 +1,8 @@
+package common
+
+type EmptyWriter struct {
```
andrewemeryanz: This could probably just live in `server.go` and not be exported:
```go
type emptyWriter struct {
    ...
}
```

andrewemeryanz (changes requested):

core/server.go
```diff
@@ -21,6 +21,12 @@ type ServerParams struct {
 	grpcManager        GrpcManager
 	prometheusRegistry *prometheus.Registry
 }
+type emptyWriter struct {
+}
+
+func (e *emptyWriter) Write(p []byte) (n int, err error) {
+	return 0, nil
```
andrewemeryanz: This should be:
```go
return len(p), nil
```
From the docs:
> Write must return a non-nil error if it returns n < len(p)

andrewemeryanz (changes requested):
Minor change. Otherwise fine.

andrewemeryanz (approved):
Looks good.

Issue #136, opened 2020-06-23
Logrus log messages not logged through pkg logger


andrewemeryanz:
Logs logged against the `pkg` logger aren't forwarding on the message to `logrus`:
```
# message logged against the `pkg` logger
time="2020-06-23T10:14:38+10:00" level=info func="github.com/anz-bank/sysl-go/core.(*logrusHook).OnLogged" file="/Users/Shared/Library/go/pkg/mod/github.com/anz-bank/sysl-go@v0.39.0/core/logging.go:16"
```

The issue appears to be that logrus overwrites the entry message:
```go
func (h *logrusHook) OnLogged(entry *log.LogEntry) error {
    e := pkgLogEntryToLogrusEntry(h.logger, entry)
    e.Log(e.Level) // should be: e.Log(e.Level, e.Message)
    return nil
}
```

andrewemeryanz:
Fixed.

PR #139, opened 2020-06-23
log messages not logged through pkg logger #136

tiagralee:
- update logrus hook to add the log entry message
- update pkg logger version
- add test to test the record logger and logrus entry

### DIFF
--- a/core/server.go
+++ b/core/server.go
@@ -4,6 +4,7 @@ package core
 
 import (
 	"context"
+	"errors"
 
 	"github.com/anz-bank/pkg/log"
 	"github.com/anz-bank/sysl-go/common"
@@ -118,7 +119,9 @@ func (params *ServerParams) Start() error {
 
 	// Panic if REST&gRPC are not running
 	if !restIsRunning && !grpcIsRunning {
-		panic("Both servers are set to nil")
+		err := errors.New("REST and gRPC servers cannot both be nil")
+		log.Error(ctx, err)
+		panic(err)
 	}
 
 	errChan := make(chan error, 1)
--- a/core/server_test.go
+++ b/core/server_test.go
@@ -2,16 +2,24 @@ package core
 
 import (
 	"context"
+	"strings"
 	"testing"
 
+	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/require"
 )
 
 func TestServerStartWithoutHandlerPanic(t *testing.T) {
+	logger, hook := test.NewNullLogger()
+	logger.SetReportCaller(true)
 	serverError := make(chan error)
 
 	require.Panics(t, func() {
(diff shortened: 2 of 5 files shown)


core/server.go
```diff
@@ -118,6 +118,8 @@ func (params *ServerParams) Start() error {
 
 	// Panic if REST&gRPC are not running
 	if !restIsRunning && !grpcIsRunning {
+		err := common.CreateError(params.Ctx, common.InternalError, "cannot start the server", nil)
```
andrewemeryanz: I support the idea of logging something here but I'd restructure it slightly. `common.CreateError` is designed to be returned from a handler:
```go
err := errors.New("REST and gRPC servers cannot both be nil")
log.Error(ctx, err)
panic(err)
```

core/server_test.go
```diff
@@ -2,16 +2,24 @@ package core
...
 )
 
 func TestServerStartWithoutHandlerPanic(t *testing.T) {
+	logger, hook := test.NewNullLogger()
+	logger.SetReportCaller(true)
 	serverError := make(chan error)
 
 	require.Panics(t, func() {
-		err := Server(context.Background(), "test", nil, nil, nil, nil)
+		err := Server(context.Background(), "test", nil, nil, logger, nil)
```
andrewemeryanz: `Server` is deprecated. While you're in the area, can you refactor this to use `ServerParams`.

andrewemeryanz (changes requested):
Minor changes requested.

core/server_test.go
```diff
@@ -2,16 +2,29 @@ package core
...
+	logger, hook := test.NewNullLogger()
+	logger.SetReportCaller(true)
 	serverError := make(chan error)
 
 	require.Panics(t, func() {
-		err := Server(context.Background(), "test", nil, nil, nil, nil)
+		err := NewServerParams(context.Background(), "test",
+			WithPkgLogger(),
+			WithLogrusLogger(logger),
+			WithRestManager(nil),
```
andrewemeryanz: You can leave out the `nil` lines:
```go
NewServerParams(context.Background(), "test", WithLogrusLogger(logger)).Start()
```

core/server.go
```diff
@@ -118,6 +119,8 @@ func (params *ServerParams) Start() error {
 
 	// Panic if REST&gRPC are not running
 	if !restIsRunning && !grpcIsRunning {
+		err := errors.New("REST and gRPC servers cannot both be nil")
+		log.Error(ctx, err)
 		panic("Both servers are set to nil")
```
andrewemeryanz: `panic(err)`

andrewemeryanz (changes requested):

andrewemeryanz (approved):
Good.

PR #140, opened 2020-06-23
Replace use of deprecated if/else with cond

orlade-anz:
This will require a relatively recent version of arr.ai, since the cond syntax changed a couple of times.


orlade-anz:
Subsumed by https://github.com/anz-bank/sysl-go/pull/142

PR #145, opened 2020-06-24
Enforce json validation on required fields when decoding.

### LINKED ISSUE #108: Enforce JSON required fields when decoding
`sysl-go` automatically decodes an incoming request body in the following manner:
```go
decoder := json.NewDecoder(r.Body)
decodeErr := decoder.Decode(&req.Request)
if decodeErr != nil {
    common.HandleError(ctx, w, common.BadRequestError, "Error reading request body", decodeErr, s.genCallback.MapError)
    return
}
```

Given the following example describes a request body with required `name` and `dateOfBirth` values and an optional `nickname` value:
```go
type RegisterRequest struct {
    Name         string            `json:"name"`
    Nickname     *string           `json:"nickname"`
    DateOfBirth  date.Date         `json:"dateOfBirth"`
}
```

If the client passes an empty `json` object: `{}`, the `json` decoder will construct the following equivalent object:
```go
RegisterRequest{
    Name: "",
    Nickname: nil,
    DateOfBirth: date.Date{},
}
```

The decoded value is then passed to the request handler to process. In some instances it's sufficient to check for the default value of a field and reject it, however in some instances it's necessary to differentiate between a missing or `null` value and the default value. In the example above, the default `date.Date` object is equivalent to `1970-01-01`, making it impossible to know if the object included a `dateOfBirth` value or not. Ideally, an error should be encountered if a `json` object does not contain a required value.

The `json` decoder has poor support for checking required values in decoding, however [this](https://stackoverflow.com/a/39134081) approach describes how it can be done. The basic approach involves generating an equivalent type with fields of pointer types and manually checking their nil values:

```go
type registerRequest_ struct {
    Name        *string    `json:"name"`
    Nickname    *string   `json:"nickname"`
    DateOfBirth *date.Date `json:"dateOfBirth"`
}

func (target *RegisterRequest) UnmarshalJSON(data []byte) error {
    inner := registerRequest_{}
    err := json.Unmarshal(data, &inner)
    if err != nil {
        return err
    }
    if inner.Name == nil {
        return errors.New("name cannot be nil")
    }
    if inner.DateOfBirth == nil {
        return errors.New("dateOfBirth cannot be nil")
    }
    *target = RegisterRequest{
        Name:        *inner.Name,
        Nickname:    inner.Nickname,
        DateOfBirth: *inner.DateOfBirth,
    }
    return nil
}
```

springwiz:
Introduces not null validation of json fields in an incoming request body.

1.  Customise demarshalling of the request body including null checks.
2. Arrai transform changes.

Fixes #108


codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
                         >> $`${go.name(.@item('name').s)} ${go.type(.@item('type'))}`
                     ::\i}
                 }
+
+                ${
+                let opParams = ep('restParams')?('queryParam')?.a:{} where $`${sysl.isPtr(.@item('type'))}`;
```
anzdaddy: Indent the body relative to `${`.

anzdaddy: No need for the expression-string wrapper.
```suggestion
                let opParams = ep('restParams')?('queryParam')?.a:{} where sysl.isPtr(.@item('type'));
```

codegen/arrai/sysl.arrai
```diff
@@ -5,6 +5,8 @@ let app = (
         ((app('endpoints') where .@value('name').s = epname) single).@value,
 );
 
+let isPtr = \t cond {t('opt')?:false && !({"sequence", "set"} & (t => .@)): true};
```
anzdaddy:
```suggestion
let isPtr = \t t('opt')?:false && !({"sequence", "set"} & (t => .@));
```

codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
...
+                cond {opParams: $`
```
anzdaddy: This will error if it doesn't match. `opParams` is treated as a name to bind, not a value to test against and fails if the existing binding is different. Use `(opParams)` instead.

codegen/arrai/svc_types.arrai
anzdaddy: Prefer the name `optParams`. It's less ambiguous than `opParams` (operation params?)

codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
...
+                    func (r *${typename}) UnmarshalJSON(data []byte) error {
+                        inner := ${typename}{}
+                        err := json.Unmarshal(data, &inner)
+                        if err != nil {
```
anzdaddy: Fold into previous line.

codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
                         >> $`${go.name(.@item('name').s)} ${go.type(.@item('type'))}`
                     ::\i}
                 }
+
+                ${
+                let opParams = ep('restParams')?('queryParam')?.a:{} where $`${sysl.isPtr(.@item('type'))}`;
+                cond {opParams: $`
+                    func (r *${typename}) UnmarshalJSON(data []byte) error {
+                        inner := ${typename}{}
```
anzdaddy:
```suggestion
                        var inner ${typename}
```

codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
...
+                        err := json.Unmarshal(data, &inner)
+                        if err != nil {
+                            return err
+                        }
+                    ${ep('restParams')?:{}
+                        -> (.('queryParam')?.a:{}) ++ (.('urlParam')?.a:{})
+                        where $`${sysl.isPtr(.@item('type'))}` orderby sysl.source(.@item('type'))
```
anzdaddy:
```suggestion
                        where sysl.isPtr(.@item('type')) orderby sysl.source(.@item('type'))
```

codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
...
+                        >> $`
+                            if inner.${go.name(.@item('name').s)} == nil {
+                                return errors.New("${go.name(.@item('name').s)} cannot be nil")
+                            }
+                        `
```
anzdaddy:
```suggestion
                        >>
                            let name = go.name(.@item('name').s);
                            $`
                                if inner.${name} == nil {
                                    return errors.New("${name} cannot be nil")
                                }
                            `
```

codegen/arrai/svc_types.arrai
```diff
@@ -69,6 +69,37 @@ let orderedTypes = \types
...
+                    ::\i}
+                        *r = ${typename}{
+                            ${ep('restParams')?:{}
+                                -> (.('queryParam')?.a:{}) ++ (.('urlParam')?.a:{}) orderby sysl.source(.@item('type'))
+                                >> $`${go.name(.@item('name').s)}: inner.${go.name(.@item('name').s)},`
```
anzdaddy:
```suggestion
                                >> let name = go.name(.@item('name').s); $`${name}: inner.${name},`
```

anzdaddy (changes requested):

codegen/arrai/svc_types.arrai
springwiz: json.Unmarshal will error out, if inner is nil.

codegen/arrai/svc_types.arrai
springwiz: This change breaks with
panic: misshapen node AST: (
  @choice: [0, 6],
  cdelim: ('': 3240‣")"),
  identpattern: (IDENT: ('': 3231‣optParams)),
  odelim: ('': 3230‣"("),
)

codegen/arrai/svc_types.arrai
springwiz: ok

springwiz: ok

codegen/arrai/sysl.arrai
springwiz: ok

codegen/arrai/svc_types.arrai
springwiz: ok

codegen/arrai/svc_types.arrai
springwiz: ok

codegen/arrai/svc_types.arrai
springwiz: ok

codegen/arrai/svc_types.arrai
anzdaddy: I think this is fixed in the current version of arr.ai.

codegen/arrai/svc_types_extn.arrai
```diff
@@ -0,0 +1,44 @@
...
+                                $`Request: inner.Request,`
+                            ::\i}
+                        }
+                        return nil
+                    }
+                `}}
+            `
+        ::\i}
+
+    `
```
anzdaddy: Missing \n.

codegen/arrai/svc_types_extn.arrai
```diff
@@ -0,0 +1,44 @@
...
+                                >> let name = go.name(.@item('name').s); $`${name}: inner.${name},`
+                            ::\i}
+                            ${(ep('param')?.a:{}) where "body" <: sysl.patterns(.@item('type')) >>
+                                $`Request: inner.Request,`
+                            ::\i}
+                        }
+                        return nil
+                    }
+                `}}
+            `
```
anzdaddy: Matching delimiters should be on lines that are at the same indent level:
```suggestion
            $`${
                let optParams = ep('restParams')?('queryParam')?.a:{} where sysl.isPtr(.@item('type'));
                cond {optParams: $`
                    func (r *${typename}) UnmarshalJSON(data []byte) error {
                        inner := ${typename}{}
                        err := json.Unmarshal(data, &inner)
                        if err != nil {
                            return err
                        }
                        ${ep('restParams')?:{}
                            -> (.('queryParam')?.a:{}) ++ (.('urlParam')?.a:{})
                            where sysl.isPtr(.@item('type')) orderby sysl.source(.@item('type'))
                            >> 
                                let name = go.name(.@item('name').s);
                                $`
                                    if inner.${name} == nil {
                                        return errors.New("${name} cannot be nil")
                                    }
                                `
                        ::\i}
                        *r = ${typename}{
                            ${ep('restParams')?:{}
                                -> (.('queryParam')?.a:{}) ++ (.('urlParam')?.a:{}) orderby sysl.source(.@item('type'))
                                >> let name = go.name(.@item('name').s); $`${name}: inner.${name},`
                            ::\i}
                            ${(ep('param')?.a:{}) where "body" <: sysl.patterns(.@item('type')) >>
                                $`Request: inner.Request,`
                            ::\i}
                        }
                        return nil
                    }
                `}
            }`
```

anzdaddy (approved):
Approved, conditional on making the changes below.

codegen/arrai/svc_types_extn.arrai
springwiz: ok

codegen/arrai/svc_types_extn.arrai
springwiz: Added

PR #147, opened 2020-06-28
Clean up arr.ai GRPC Client codegen

anzdaddy:


nydrani (approved):
LGTM

NOTE: breaking change in Client struct in codegen/arrai/grpc_client.arrai (renamed field)

Issue #133, opened 2020-06-22
Identify and resolve gaps between `sysl` transforms and `arrai` transforms.


springwiz:
@springwiz Can you close this ticket and instead create a separate ticket to identify all gaps between `sysl` transforms and `arrai` transforms.

_Originally posted by @andrewemeryanz in https://github.com/anz-bank/sysl-go/issues/121#issuecomment-647253848_

andrewemeryanz:
@springwiz I'll attempt to migrate the first service to get a benchmark of code comparison. I'll then hand this over to you to investigate the remaining services.

ghost:
@andrewemeryanz you probably want to check on my changes in https://github.service.anz/dcx/anzapp-bff/tree/master/codegen/arrai asI've already fixed a bunch of issues with the rest code

andrewemeryanz:
@anz-gordonj7 Thanks.

springwiz:
@anz-gordonj7 Can you point me to issues that you have fixed, in the arrai code included in dcx/anzapp-bff.

ghost:
diff the dcx/codegen/arrai directory with sysl-go/codegen/arrai. I don't have a list of changes made

andrewemeryanz:
Migration complete.

PR #149, opened 2020-06-29
Begin moving sysl-go sample code gen to arrai.

springwiz:
Move sysl-go sample code generation from sysl to arrai.

Step 1: Move Types.
Step 2: Move rest of the stuff.


codegen/tests/simple/types.go
```diff
@@ -112,6 +114,33 @@ type GetStuffListRequest struct {
 	It int64
 }
 
+func (r *GetStuffListRequest) UnmarshalJSON(data []byte) error {
+	inner := GetStuffListRequest{}
```
andrewemeryanz, line 151: This generated code is incorrect because:
1. It results in an infinite loop when unmarshalling, and
1. It's returning an error if an optional value is `nil`.

The correct approach is to:
1. Create a separate (inner) struct where all fields are pointers
1. Unmarshall into the inner struct
1. Check that required fields aren't `nil`
1. Construct a new target struct with the unmarshalled values from the inner struct.

```go
type registerRequest_ struct {
    Name        *string    `json:"name"`
    Nickname    *string   `json:"nickname"`
    DateOfBirth *date.Date `json:"dateOfBirth"`
}

func (target *RegisterRequest) UnmarshalJSON(data []byte) error {
    inner := registerRequest_{}
    err := json.Unmarshal(data, &inner)
    if err != nil {
        return err
    }
    if inner.Name == nil {
        return errors.New("name cannot be nil")
    }
    if inner.DateOfBirth == nil {
        return errors.New("dateOfBirth cannot be nil")
    }
    *target = RegisterRequest{
        Name:        *inner.Name,
        Nickname:    inner.Nickname,
        DateOfBirth: *inner.DateOfBirth,
    }
    return nil
}
```

codegen/tests/simple/simple_test.go
```diff
@@ -398,6 +399,26 @@ func validQueryParamTest(t *testing.T, req GetStuffListRequest, query string) {
 	require.NoError(t, err)
 }
 
+func TestJsonValidateInput(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Test request parameters
+		decoder := json.NewDecoder(r.Body)
+		req := new(GetStuffListRequest)
+		decodeErr := decoder.Decode(req)
```
andrewemeryanz, line 407: This line should result in a call through to the `UnmarshallJSON` method on `GetStuffListRequest`. I did a quick test but couldn't diagnose the issue.

codegen/arrai/service.arrai
```diff
@@ -5,11 +5,12 @@ let sysl = //{./sysl};
...
 # codegen groups
 
-let restClient = {"error_types", "service", "types", "types_extn"};
-let restService = restClient | {"requestrouter", "servicehandler", "serviceinterface", "types_extn"};
+let restClient = {"error_types", "service", "types"};
+let restService = restClient | {"requestrouter", "servicehandler", "serviceinterface"};
 
 let grpcClient = {"grpc_client"};
 let grpcService = grpcClient | {"grpc_handler", "grpc_interface"};
+let migrate = {"types"};
```
andrewemeryanz: I don't think `migrate` belongs here. `service.arrai` is designed to provide a simple mechanism for users to generate a set of files for given use cases. You could potentially derive a similar file for use within the test suite itself.

andrewemeryanz (changes requested):
This change is going to be held up by the json unmarshalling work. It might be worthwhile to address this first then rework this pull request.

codegen/arrai/service.arrai
springwiz: I could principally have another copy of service.arrai within the test suite itself. But it would be similar to the existing service.arrai. It is better to have a single source of truth, otherwise there is a chance that the two copies of service.arrai would drift away from each other. We will endup merging the changes from one file into another.

codegen/tests/simple/simple_test.go
springwiz: will have a look.

springwiz: Commented on the change here https://github.com/anz-bank/sysl-go/issues/108#issuecomment-651595445. Please review.

springwiz:
> This change is going to be held up by the json unmarshalling work. It might be worthwhile to address this first then rework this pull request.

JSON Marshalling changes moved to PR https://github.com/anz-bank/sysl-go/pull/153.

springwiz:
Part of #133 issue resolution.

codegen/arrai/svc_types.arrai
```diff
@@ -105,6 +105,36 @@ let orderedTypes = \types
                         >> $`${go.name(.@item('name').s)} ${go.type(.@item('type'))}`
                     ::\i}
                 }
+            ${
```
andrewemeryanz, line 108: This looks like a duplicate to the json validation already merged.

andrewemeryanz (changes requested):
Duplicate section. Otherwise fine.

codegen/arrai/svc_types.arrai
springwiz: ya that would be resolved during merge.

andrewemeryanz: This is a pull request to merge to master. If this file shouldn't be changed from what master is presently then this file wouldn't show any changes.

springwiz:
Opening another PR for the same.

PR #151, opened 2020-07-03
Attach verbose logging level to request context

### LINKED ISSUE #150: log is not being populate with right context
Hi team,

after upgrade to version 0.47, we are see logs are not being printed as expected:

```
2020-07-02T03:56:27.966988002Z latency=412119589 Downsteam=AAA X-B3-Spanid=XXXXX status=400 traceid=XXXX method=POST X-B3-ParentSpanId=XXXX X-B3-Traceid=XXXXX request=https://BBBBBBBBB remote= took=412.119589ms INFO Backend request completed 
```
We dont have the "Cause" for this 400 code any more. And the Log Level is showing "INFO" here which should not be.

Also, one of the important thing that, notice that "Request Header" and "Response Header" are not longer in place as well. is there a reason for that? i think for dev and sit environment we will need this "verbose" output.

andrewemeryanz:
Internally, sysl-go uses a value attached to the context to
track whether or not additional debug/verbose-level information is
attached to its logs. In order for this value to persist within a
request, it must also be attached to the request context.

Closes #150

### DIFF
--- a/logconfig/logconfig.go
+++ b/logconfig/logconfig.go
@@ -1,0 +1,27 @@
+package logconfig
+
+import (
+	"context"
+)
+
+type isVerboseLoggingKey struct{}
+type isVerboseLogging struct {
+	Flag bool
+}
+
+// Get whether or not sysl-go internals log additional debug/verbose level information.
+func IsVerboseLogging(ctx context.Context) bool {
+	v, ok := ctx.Value(isVerboseLoggingKey{}).(*isVerboseLogging)
+	if ok {
+		return v.Flag
+	}
+	return false
+}
+
+// Set against the context whether or not sysl-go internals log additional debug/verbose level information.
+func SetVerboseLogging(ctx context.Context, verbose bool) context.Context {
+	return context.WithValue(ctx, isVerboseLoggingKey{},
+		&isVerboseLogging{
+			Flag: verbose,
+		})
+}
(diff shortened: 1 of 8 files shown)


logconfig/logconfig.go
```diff
@@ -0,0 +1,30 @@
...
+	v, ok := ctx.Value(isVerboseLoggingKey{}).(*isVerboseLogging)
+	if ok {
+		return v.Flag
+	}
+	return false
+}
+
+// Set against the context whether or not sysl-go internals log additional debug/verbose level information.
+func SetVerboseLogging(ctx context.Context, verbose bool) context.Context {
+	ctx = log.WithConfigs(log.SetVerboseMode(verbose)).Onto(ctx)
```
tiagralee: set the verbose mode to context as well, otherwise, the `field.Debug` cannot be logged.

andrewemeryanz: I don't know if this is required. The `pkg` logger stored against the old context should already be `verbose`. If the state isn't being transferred correctly then you might want to dig into the interaction between `log.From` and `log.WithLogger`.

andrewemeryanz (commented):
Looks good.

tiagralee:
LGTM

tiagralee (approved):
LGTM

uynap:
Thanks Andrew & Alex. I will test it.

PR #153, opened 2020-07-06
Enforce json required

springwiz:
Build json validation capabilities in sysl-go so that its possible to validate missing or null value. It attempts to distinguish between default values that have been passed and the missing values. types.go generation is moved to arrai as a side effect of the above change.

1. Implement marshalljson for types described in sysl. Generate validation for fields which are required.
2. arrai transform changes.
3. Test cases.
4. Move types.go generation from sysl transforms to arrai.


codegen/tests/simple/types.go
```diff
@@ -23,6 +25,39 @@ type Item struct {
...
+// PostRequest ...
+type PostRequest struct {
+	Bt *bool            `json:"Bt,omitempty"`
+	Dt convert.JSONTime `json:"Dt"`
+	St string           `json:"St"`
+}
+
+func (t *PostRequest) UnmarshalJSON(data []byte) error {
+	inner := struct {
+		Bt *bool
```
andrewemeryanz: I suspect these require struct tags for json also:
```go
Bt *bool            `json:"Bt,omitempty"`
```

codegen/tests/simple/types.go
```diff
@@ -23,6 +25,39 @@ type Item struct {
...
+	if inner.Dt == nil {
+		return errors.New("Dt cannot be nil")
+	}
+
+	if inner.St == nil {
+		return errors.New("St cannot be nil")
+	}
+
+	*t = PostRequest{
+		Dt: *inner.Dt,
```
andrewemeryanz, line 55: These should be ordered in the same manner as above.

codegen/arrai/service.arrai
```diff
@@ -10,6 +10,7 @@ let restService = restClient | {"requestrouter", "servicehandler", "serviceinter
 
 let grpcClient = {"grpc_client"};
 let grpcService = grpcClient | {"grpc_handler", "grpc_interface"};
+let migrate = {"types"};
```
andrewemeryanz, line 13: I don't think this is necessary. You can just pass through `types.go`.

andrewemeryanz (changes requested):

codegen/arrai/service.arrai
springwiz: This is part of bigger change where we move more files, anyways they have to move to arrai sooner or later.

codegen/tests/simple/types.go
springwiz: That shouldn't affect the code behaviour but ok.

codegen/tests/simple/types.go
springwiz: ok

codegen/arrai/service.arrai
andrewemeryanz: We'll migrate all generated code in one go, not one generated file at a time. This is a temporary measure to get some confidence that the generated json code is correct.

springwiz: There is another PR open which migrates everything else, with migrate as target. We can leave this here, it will change anyway once you merge that PR.

andrewemeryanz (approved):

PR #157, opened 2020-07-10
Grpc client injection

anzdaddy:


joshcarp (approved):
Don't really understand what's going on here, but hopefully will soon

Issue #159, opened 2020-07-14
Is there a way to retrieve the error messages from downstreams?


uynap:
For example, UI -> BFF -> OtherAPI, when OtherAPI response:
```
< HTTP/1.1 500 Server Error
{
  "message" : [ {
    "text" : "Error while making call to ABC.",
    "type" : "error"
  } ]
}
```
At the moment, the BFF will response UI:
```
< HTTP/1.1 503 Server Error
```
without any details.

I know the raw message from the OtherAPI can be logged in sysl-go, but it’s hard to access the OtherAPI's customised error messages in BFF's handler functions. There are some use cases that the UI needs to present different things base on the error message received from the OtherAPI.

tiagralee:
Linked Issue https://github.com/anz-bank/sysl-go/issues/105

andrewemeryanz:
Implemented in https://github.com/anz-bank/sysl-go/pull/188.
Pattern of use:
```go
ctx = common.ProvisionRestResult(ctx)  // provision for the result in the context
response, err := client.PostRestFooBarWithArg(ctx, &request)
result := common.GetRestResult(ctx)  // retrieve the result from the context
```

PR #160, opened 2020-07-14
Include RestRequestOption interface and OnResult implementation

### LINKED ISSUE #105: Need to be able to pull headers out of the restlib response

andrewemeryanz:
Custom handler implementations in some instances require access to
more information than downstream service calls currently permit. This
can include the status code, headers or body in the instance that a
non-successful response was returned.

RestRequestOption instances can be passed into downstream REST
requests to perform custom actions:

```go
response, err := client.GetRestFooBar(ctx, &request,
    restlib.RestRequestOnResult(func(result *restlib.HTTPResult, err error) {
        // access to raw http result and error
    }))
```

Closes #105 #159

### DIFF
--- a/restlib/restlib.go
+++ b/restlib/restlib.go
@@ -188,3 +188,29 @@ func SetHeaders(w http.ResponseWriter, headerMap http.Header) {
 		}
 	}
 }
+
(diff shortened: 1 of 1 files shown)


andrewemeryanz:
Closing to reopen as draft #160

Issue #164, opened 2020-07-16
Render image content type for Upstream


suipinglu:
Hi everyone,

BFF needs to support serving images for UI, currently having issues for rendering that.

[The request for downstream does not seem like support content type image for the moment as well](https://github.com/anz-bank/sysl-go/blob/master/restlib/restlib.go#L82)

Also, seems like the reponse handling need some supports as well:
[Need 'image' for contentType support](https://github.com/anz-bank/sysl-go/blob/master/restlib/restlib.go#L39)

Cheers,
Ping

PR #168, opened 2020-07-20
Render image content type for Upstream #164

tiagralee:
- update restlib to support image content-type
- add related tests

### DIFF
--- a/restlib/restlib.go
+++ b/restlib/restlib.go
@@ -49,14 +49,17 @@ func unmarshal(resp *http.Response, body []byte, respStruct interface{}) (*HTTPR
 		return makeHTTPResult(resp, body, nil), nil
 	}
 
+	contentType := resp.Header.Get("Content-Type")
+
 	e := reflect.ValueOf(respStruct).Elem()
-	if e.Kind() == reflect.String {
+	kind := e.Kind()
+
+	if kind == reflect.String || (kind == reflect.Slice && e.Type().Elem().Name() == "uint8") {
 		p := reflect.New(e.Type())
 		p.Elem().Set(reflect.ValueOf(body).Convert(e.Type()))
 		return makeHTTPResult(resp, body, p.Interface()), nil
 	}
 
-	contentType := resp.Header.Get("Content-Type")
 	if strings.Contains(contentType, "xml") {
 		respStruct = string(body)
 	} else {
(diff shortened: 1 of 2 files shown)


restlib/restlib.go
```diff
@@ -57,7 +57,7 @@ func unmarshal(resp *http.Response, body []byte, respStruct interface{}) (*HTTPR
 	}
 
 	contentType := resp.Header.Get("Content-Type")
-	if strings.Contains(contentType, "xml") {
+	if strings.Contains(contentType, "xml") || strings.Contains(contentType, "image") {
```
andrewemeryanz: @tiagralee Can you check with @suipinglu that this solves his issue.

His issue tags a different line than the one touched in this pull request:
https://github.com/anz-bank/sysl-go/blob/master/restlib/restlib.go#L82

andrewemeryanz (changes requested):
Check with owner of issue and add `Closes #164` to the commit message.

restlib/restlib_test.go
```diff
@@ -65,6 +66,19 @@ func TestUnmarshalEmptyBodyOK(t *testing.T) {
 	require.Nil(t, result.Response)
 }
 
+func TestUnmarshalImageContentType(t *testing.T) {
+	header := map[string][]string{
+		"Content-Type": {"image/png"},
+	}
+	result, err := unmarshal(&http.Response{Header: header}, []byte(okJSON), &OkType{})
```
andrewemeryanz: The logic is fine but can you change the example from: 
```
[]byte(okJSON)
``` 

into something like 
```
var image = [...]byte { 0, 1 }
```

I was a bit confused when the test is for an image response and the response contained a json string.

andrewemeryanz (changes requested):

restlib/restlib.go
```diff
@@ -57,9 +58,12 @@ func unmarshal(resp *http.Response, body []byte, respStruct interface{}) (*HTTPR
 	}
 
 	contentType := resp.Header.Get("Content-Type")
-	if strings.Contains(contentType, "xml") {
+	switch {
+	case strings.Contains(contentType, "xml"):
 		respStruct = string(body)
-	} else {
+	case strings.Contains(contentType, "image"):
+		respStruct = base64.StdEncoding.EncodeToString(body)
```
andrewemeryanz: We shouldn't be Base64 encoding this. A byte array should suffice.

andrewemeryanz (changes requested):

restlib/restlib.go
```diff
@@ -49,14 +49,16 @@ func unmarshal(resp *http.Response, body []byte, respStruct interface{}) (*HTTPR
 		return makeHTTPResult(resp, body, nil), nil
 	}
 
+	contentType := resp.Header.Get("Content-Type")
+
 	e := reflect.ValueOf(respStruct).Elem()
-	if e.Kind() == reflect.String {
+	kind := e.Kind()
+	if kind == reflect.String || (kind == reflect.Slice && strings.Contains(contentType, "image")) {
```
andrewemeryanz: Is it possible to generalise this check to byte arrays? 
```
Current:  Is this a slice and an image.
Proposed: Is this a slice of bytes.
```

tiagralee: Good catch, the "text/plain" for ApiDoc is slice as well, I'll have a look if it can further check the type in the slice.

andrewemeryanz (approved):
Great.

PR #171, opened 2020-07-20
Generate DB connection based on DB config

orlade-anz:
sysl-go's current DB codegen is hardcoded to produce a connection to an in-memory sqlite database with a hardcoded schema. This PR parameterises the codegen with DB config such that the connection created will be to the specified DB.


codegen/arrai/sysl.arrai
```diff
@@ -1,8 +1,20 @@
...
+
+    # dbConfig returns a tuple of DB config parameters specified as app attributes.
+    dbConfig: \app cond patterns(app) {{"DB", ...}: (
+        type: app('attrs')('db_type')?.@value.s:'',
+        name: app('attrs')('db_name')?.@value.s:'',
+        user: app('attrs')('db_user')?.@value.s:'',
+        password: app('attrs')('db_password')?.@value.s:'',
+        host: app('attrs')('db_host')?.@value.s:'',
+        port: app('attrs')('db_port')?.@value.s:'',
+    )},
```
anzdaddy, line 17: Runtime config is passed through the generated app's config file, not via Sysl definitions.

anzdaddy (changes requested):

andrewemeryanz:
Dormant.

PR #173, opened 2020-07-21
Fix generation using arrai

joshcarp:
The following problems presented with https://github.com/anz-bank/sysl-template

```go
// Service interface for simple
type Service interface {
	Get(ctx context.Context, req *GetRequest) (*Welcome, error)
	GetFoobarList(ctx context.Context, req *GetFoobarListRequest) (*jsonplaceholder.todosResponse, error)
}
```
There is a reference to an unexported field, that should be exported; ie `*jsonplaceholder.TodosResponse` instead of `*jsonplaceholder.todosResponse`

This is the fix to make it work

This isn't to be merged in immediately, as there is some code gen that relies on the current version of arrai and these changes might be breaking changes

### DIFF
--- a/codegen/arrai/go.arrai
+++ b/codegen/arrai/go.arrai
@@ -125,7 +125,7 @@ let module = \module
         let pkgnames = type => .@item;
         let packages = module('apps') where .@ <: pkgnames;
         cond {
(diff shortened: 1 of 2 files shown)


marcelocantos (approved):

andrewemeryanz (approved):

joshcarp:
Doesn't seem to impact other repos which I tested locally

PR #174, opened 2020-07-21
Update default callback to implement core.RestGenCallback interface

joshcarp:
Updates the default callback to implement interface defined [here](https://github.com/anz-bank/sysl-go/blob/eedeba910dcdb036374e966dab0d53f1cc5fbc39/core/callback.go#L11)

### DIFF
--- a/common/defaultcallback.go
+++ b/common/defaultcallback.go
@@ -53,3 +53,7 @@ func (g Callback) HandleError(ctx context.Context, w http.ResponseWriter, kind K
 func (g Callback) DownstreamTimeoutContext(ctx context.Context) (context.Context, context.CancelFunc) {
 	return context.WithTimeout(ctx, g.DownstreamTimeout)
 }
+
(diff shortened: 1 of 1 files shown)


common/defaultcallback.go
```diff
@@ -53,3 +53,7 @@ func (g Callback) HandleError(ctx context.Context, w http.ResponseWriter, kind K
 func (g Callback) DownstreamTimeoutContext(ctx context.Context) (context.Context, context.CancelFunc) {
 	return context.WithTimeout(ctx, g.DownstreamTimeout)
 }
+
+func (g Callback) MapError(ctx context.Context, err error) *HTTPError {
```
orlade-anz, line 54: Should this be used above on line 48?

`httpError := MapError(ctx, se)`

joshcarp:
Yeah that's a lot more useful

orlade-anz (approved):

PR #178, opened 2020-07-21
Feature/docker

joshcarp:
- Adds dockerfile to build with sysl, arrai and sysl-go for completeness
(sysl might be able to be left out of this image)
- Also adds a github action that builds and pushes to dockerhub.com/r/anzbank/sysl-go


orlade-anz (approved):
LGTM, but please add some reference to usage of the Docker image.

PR #179, opened 2020-07-22
Fix generation of sysl.json files in testdata

### LINKED ISSUE #177: testdata sysl.json files are not regenerated by make
https://github.com/anz-bank/sysl-go/pull/170 appears to have broken sysl-go since it regenerated and committed test outputs but not `testdata/simple/sysl.json`.

This would be fine if `make` caused `testdata/simple/sysl.json` to be regenerated before regenerating the derived code, but it doesn't. Thus the old code from before https://github.com/anz-bank/sysl-go/pull/170 is generated and the tests fail.

orlade-anz:
Fixes #177 

Currently make doesn't regenerate `testdata/*/sysl.json` from `testdata/*/*.sysl`, meaning tests don't necessarily run against the latest generated code following changes to the Sysl specs.

This solution is not good Makefile style, but it fixes the problem. The whole make system in sysl-go needs an overhaul, so I'm not keen to take on the whole thing for this fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,8 @@ SYSL_GO_ROOT=github.com/anz-bank/sysl-go
 MIGRATE=migrate
 
 define run-arrai
+$(eval NAME := $(shell echo $< | tr '[:upper:]' '[:lower:]'))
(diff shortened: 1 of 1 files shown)


springwiz (approved):
LGTM.

PR #187, opened 2020-07-24
Update Docker to format after code generation

joshcarp:
Adds a arrai-docker script that formats code after code generation


scripts/arrai-docker.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+$@ | tar xf - -C /out/
+go get -u github.com/rickb777/date
```
orlade-anz: Nit: This is an odd requirement... I'd expect something more like `go get -u golang.org/x/tools/cmd/goimports` to update its transitive dependencies.

If this is necessary, can you add a comment briefly describing why? Is it something specific to sysl-catalog?

joshcarp: oh yeah, the date one seems to be a weird one where in a fresh environment the date package isn't imported by goimports so this is explicitly requiring it
I haven't done enough investigation into why it's only this package

orlade-anz: That's fine; maybe just add a comment, or refer to an issue to figure it out.

orlade-anz (approved):

PR #188, opened 2020-07-24
ProvisionRestResult implementation

### LINKED ISSUE #105: Need to be able to pull headers out of the restlib response

tiagralee:
Custom handler implementations in some instances require access to
more information than downstream service calls currently permit. This
can include the status code, headers or body in the instance that a
non-successful response was returned.

Custom code can ProvisionRestResult and retrive the RestResult after the downstream call
Generated code will set the HTTP result as the RestResult

close #105 #159

### DIFF
--- a/common/corerequestcontext.go
+++ b/common/corerequestcontext.go
@@ -49,6 +49,12 @@ type respHeaderAndStatusContext struct {
 	status int
 }
 
+type RestResult struct {
+	StatusCode int
+	Headers    map[string][]string
+	Body       []byte
+}
+
 // LoggerToContext create a new context containing the logger
 // Deprecated: Use ServerParams.WithPkgLogger instead
 func LoggerToContext(ctx context.Context, logger *logrus.Logger, entry *logrus.Entry) context.Context {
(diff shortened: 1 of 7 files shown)


common/corerequestcontext.go
```diff
@@ -171,3 +177,21 @@ func (t *tempRoundtripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	ctx := log.With("Downsteam", t.name).Onto(r.Context())
 	return internal.NewLoggingRoundTripper(ctx, t.base).RoundTrip(r)
 }
+
+type RestResultContextKey struct{}
+
+// Provision within the context the ability to retrieve the result of a rest request
+func ProvisionRestResult(ctx context.Context) context.Context {
+	// insert an object in the context
```
andrewemeryanz: Remove this comment:
```go
// insert an object in the context
```

restlib/restlib.go
```diff
@@ -193,3 +193,29 @@ func SetHeaders(w http.ResponseWriter, headerMap http.Header) {
...
+		StatusCode: result.HTTPResponse.StatusCode,
+		Headers:    result.HTTPResponse.Header,
+		Body:       result.Body,
+	}
+	OnRestResult(ctx, restResult, err)
+}
+
+// OnRestResult method will be called from tests as we don't expose the HTTPResult itself
+func OnRestResult(ctx context.Context, result *common.RestResult, err error) {
+	// find the object in the context (if available) and set the result
```
andrewemeryanz: Remove this comment:
```go
// find the object in the context (if available) and set the result
```

restlib/restlib.go
```diff
@@ -193,3 +193,29 @@ func SetHeaders(w http.ResponseWriter, headerMap http.Header) {
 		}
 	}
 }
+
+// OnRestResultHTTPResult is called from generated code when an HTTP result is retrieved.
+func OnRestResultHTTPResult(ctx context.Context, result *HTTPResult, err error) {
+	if result == nil || result.HTTPResponse == nil {
```
andrewemeryanz: This logic isn't correct. If we only have an error then we want to propagate that also:
```go
var restResult *common.RestResult = nil
if result != nil && result.HTTPResponse != nil {
    restResult = ...
}
OnRestResult(ctx, restResult, err)
```

andrewemeryanz (changes requested):
Great. Couple of minor changes.

andrewemeryanz (approved):
Perfect.

PR #192, opened 2020-07-28
Cleanup linter diagnostics

anzdaddy:
Mostly disable a bunch of new linters, probably due to running a newer version of golangci-lint.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,20 +6,58 @@ linters-settings:
     line-length: 180
 
 linters:
-  enable-all: true
-  disable:
-    - dupl
-    - gochecknoglobals
-    - godot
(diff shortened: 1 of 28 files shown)


.golangci.yml
```diff
@@ -7,9 +7,16 @@ linters:
   disable:
     - dupl
     - gochecknoglobals
-    - wsl
-    - lll
+    - godot
```
cuminandpaprika, line 23: Nit: Some comments here to explain why each of these linters is too noisy helps with future evaluations, if you've got it handy/ fresh in your mind.

cuminandpaprika (approved):
LGTM

joshcarp (approved):

PR #194, opened 2020-07-29
Fixes RestResult to be returned on http failure

### LINKED ISSUE #105: Need to be able to pull headers out of the restlib response

andrewemeryanz:
Previously, the RestResult returned from common.GetRestResult was
only successfully returned on successful http requests. This change
returns the result on http failure also.

Closes #105 #159


springwiz (approved):
LGTM

Issue #95, opened 2020-05-06
Support 'oneof' keyword in OpenApi to specify schemas of API response for sysl code generation


linwumeng:
Sysl code generation doesn't support 'oneof' keyword in OpenApi.
```
paths:
  /pets:
    patch:
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Dog'
      responses:
        '200':
          description: Updated
```
Request new feature of sysl to generate code of bff and clients from OpenApi specs supporting 'oneof'.

anzdaddy:
As a quick workaround, we'll expose the request directly to the handler in the case that the top level thing is a oneof.

linwumeng:
Take https://github.com/anz-bank/sysl-go/blob/master/codegen/tests/simple/servicehandler.go as example. The generated code is
```
type Handler interface {
	GetApiDocsListHandler(w http.ResponseWriter, r *http.Request)
	... 
}
type ServiceHandler struct {
	genCallback                 core.RestGenCallback
	serviceInterface            *ServiceInterface
	depsDepsService             deps.Service
	downstreamDownstreamService downstream.Service
}
func (s *ServiceHandler) GetApiDocsListHandler(w http.ResponseWriter, r *http.Request) {
	...
	apidoc, err := s.serviceInterface.GetApiDocsList(ctx, &req, client)
	if err != nil {
		common.HandleError(ctx, w, common.DownstreamUnexpectedResponseError, "Downstream failure", err, s.genCallback.MapError)
		return
	}

	headermap, httpstatus := common.RespHeaderAndStatusFromContext(ctx)
	restlib.SetHeaders(w, headermap)
	restlib.SendHTTPResponse(w, httpstatus, apidoc)
}
```
It works well when the schema of response body is a certain one instead of 'oneOf'. In this case, `apidoc` at `apidoc, err := s.serviceInterface.GetApiDocsList(ctx, &req, client)`  is of `*[]deps.ApiDoc`.
For `oneOf`, will `apidoc` be of `interface{}` as a workround?

ghost:
@anzdaddy Any updates on this please? Its going to be blocking us very soon

andrewemeryanz:
@anz-gordonj7 Issue will be picked up within the next day or two.

tiagralee:
[...]

PR #198, opened 2020-08-14
Update codegen to support Sysl union type.

ericzhang6222:
This is a simple and quick fixing as #95 is a blocker issue, now it can transfer Sysl union type too. For example, transform
```
!type Cat:
        id <: string 
        age <: int?:
            @json_tag = "age"
        hunts <: bool?:
            @json_tag = "hunts"

!type Dog:
        id <: string
        bark <: bool?:
            @json_tag = "bark"
        breed <: string?:
            @json_tag = "breed"

!union PetA:
        Cat
        Dog
```
to
```
// PetA can be one of following types in runtime:
// Cat
// Dog
type PetA interface {
	// isPetA is identifier method
	isPetA()
}

// isPetA identifies Cat is instance of PetA
func (i Cat) isPetA() {
}

// isPetA identifies Dog is instance of PetA
func (i Dog) isPetA() {
}
```
It can be enhanced still, like adding util method to transform instance of `Pet` to `Cat` or `Dog`.


codegen/arrai/svc_types.arrai
```diff
@@ -140,4 +141,10 @@ let fieldJsonTag = \type \key
             type ${.typename} struct {
             }
         `::\i}
+        ${unions >> \(@: key, @value: value, :typename)$`
+            // ${typename} can be one of following types in runtime:
+            ${value('oneOf')('type').a >> .('typeRef')('ref')('path').a >> ('// ' + .@item.s)::\i}
```
orlade-anz: Nit: style
```suggestion
            ${value('oneOf')('type').a >> .('typeRef')('ref')('path').a >> $`// ${.@item.s}`::\i}
```

orlade-anz (changes requested):
Generally LGTM I think, but I'm not super familiar with sysl-go yet. I'd like an approval from someone who is.

codegen/tests/simple/types.go
```diff
@@ -18,6 +18,20 @@ var _ = time.Parse
 // Reference imports to suppress unused errors
 var _ = date.Parse
 
+// Cat ...
+type Cat struct {
```
andrewemeryanz, line 22: We should go a little further and make the relationship explicit:

```go
func (c Cat) isPet() {}
func (d Dog) isPet() {}
type Pet interface {
	isPet()
}
```

andrewemeryanz (changes requested):
Overall approach good. Minor change requested to make the relationship between union types explicit.

springwiz (dismissed):

springwiz (changes requested):
I see an issue lurking over here, as per my understanding, both Cat and Dog can behave as Pet during runtime as they are part of a union(both implement Pet interface {}). But essentially in the generated code the method set associated with  Pet interface {} is empty, which means even a struct like type struct Empty {} can be a Pet. Please add a method to the Pet interface {} and implement the same in the structs Cat and Dog.

ericzhang6222:
> I see an issue lurking over here, as per my understanding, both Cat and Dog can behave as Pet during runtime as they are part of a union(both implement Pet interface {}). But essentially in the generated code the method set associated with Pet interface {} is empty, which means even a struct like type struct Empty {} can be a Pet. Please add a method to the Pet interface {} and implement the same in the structs Cat and Dog.

Fixed.

springwiz (approved):
LGTM

codegen/tests/simple/types.go
```diff
@@ -200,3 +229,39 @@ type Str string
 // Empty ...
 type Empty struct {
 }
+
+// PetA can be one of following types in runtime:
+// Cat
+// Dog
+type PetA interface {
+	// IsPetA returns true if it is instance of PetA
+	IsPetA() bool
```
andrewemeryanz: Method should be `isPetA()`:
1. We don't want to export the method we just want to mark the implementations.
1. We don't need to return anything; void will suffice.

andrewemeryanz (changes requested):

andrewemeryanz (approved):

orlade-anz (approved):

PR #200, opened 2020-08-17
Update arrai grpc client/service import transform to support remote protobufs

nydrani:


springwiz (changes requested):
Please show some generated code and add test cases.

nydrani:
> Please show some generated code and add test cases.

The generated code doesn't change, it's just an update to how the code generator adds the import. It used to assume that the protobufs are always local to the project, so this is just adding support to remote protobuf code imports.

springwiz (approved):
LGTM

PR #202, opened 2020-08-19
Updated grpc client transform to generate an interface

nydrani:
Added generated interface to allow for stubs to be written against the service

### DIFF
--- a/codegen/arrai/grpc_client.arrai
+++ b/codegen/arrai/grpc_client.arrai
@@ -15,6 +15,17 @@ let go = //{./go};
     $`
         ${go.prelude(app, (clientDeps => $`${basepath}/${.import}`) | go.pbPackage(app))}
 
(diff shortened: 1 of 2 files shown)


codegen/arrai/grpc_client.arrai
```diff
@@ -15,6 +15,17 @@ let go = //{./go};
     $`
         ${go.prelude(app, (clientDeps => $`${basepath}/${.import}`) | go.pbPackage(app))}
 
+        // Service interface for ${appname}.
+        type Service interface {
```
andrewemeryanz, line 19: Can you include the generated code that changes as a result of this change.

nydrani: I think the generated code is in the .gitignore file and not added to the repo.

springwiz (approved):
LGTM

PR #205, opened 2020-08-20
Added insecure flag to grpc-client service dialing

nydrani:
Currently, no tests are added due to old tests being run against sysl transforms and are not migrated to test against the arrai transforms. Sysl transforms also do not contain a grpc-client wrapper, so there is no update in that area

### DIFF
--- a/codegen/arrai/grpc_client.arrai
+++ b/codegen/arrai/grpc_client.arrai
@@ -11,7 +11,6 @@ let go = //{./go};
         $`${name}(ctx context.Context, ${pname} *pb.${go.type(ptype)}) (*pb.${rparam}, error)`;
 
     # TODO: sysl automapping between sysl types and protobuf types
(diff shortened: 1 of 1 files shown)


ashwinsajiv (approved):

codegen/arrai/grpc_client.arrai
```diff
@@ -33,11 +31,16 @@ let go = //{./go};
         }
 
         // NewClient creates a new Client.
-        func NewClient(addr string, connTimeout time.Duration) (*Client, error) {
+        func NewClient(addr string, connTimeout time.Duration, insecure bool) (*Client, error) {
```
joshcarp, line 34: Shouldn't this be determined if the addr starts with `http` or `https`?

nydrani: I dont think gRPC dialing supports uri schemes.

PR #226, opened 2021-08-24
Use latest sysl in docker

akumria:
Completely untested, well the github action is.

However the code to extract the latest tag via jq is correct.


.github/workflows/docker.yml
```diff
@@ -19,6 +19,12 @@ jobs:
       - name: Check out code
         uses: actions/checkout@master
 
+      - name: Get SYSL version
+        run: |
+          curl -H "Accept: application/vnd.gituhb.v3+json" https://api.github.com/repos/anz-bank/sysl/releases > releases.json
+          SYSL_VERSION=`cat releases.json | jq -r '.[0] | .tag_name'`
```
andrewemeryanz, line 25: It's an interesting idea to always build against the latest Sysl version but as a general approach I wouldn't recommend it. Sysl doesn't guarantee backwards compatibility so mandating that Sysl-go can be built against the latest Sysl will hold up changes to Sysl-go until alignment is made.

andrewemeryanz (changes requested):

.github/workflows/docker.yml
akumria: The issue is that when you look at https://github.com/anz-bank/sysl-go you are just told to have "sysl" on the your path.

The current docker file is using sysl v0.258 but if you install via brew (recommend) you get sysl v0.438 (from https://sysl.io/docs/installation/)

If there is no guarantee of backwards compatibility it really feels like sysl-go and sysl need to be released in lock-step.

akumria:
Closing. I am not actively using sysl at this time.

PR #253, opened 2023-05-19
Prevent resource leaks

amitkrout:
```
$ make
[...]
golangci-lint run
common/errors_go13_test.go:45:18: response body must be closed (bodyclose)
        resp := r.Result()
                        ^
[...]
```
The linting tool has identified a potential issue with the code. It's error out that the response body obtained from r.Result() should be closed after its use to prevent resource leaks.

By using the defer statement, the resp.Body.Close() function will be executed when the surrounding function returns, ensuring the response body is always closed.


AriehSchneier:
httptest.NewRecorder().Result().Body is an io.NopCloser

